PR #4, opened 2016-07-19
Add Prout to track deployment of pull requests

rtyley:
Using the origin url at the moment, as we've not yet pointed Fastly at these servers - and `disableSSLVerification` because we haven't got the certificates sorted yet.

@AWare


AWare:
👍

PR #5, opened 2016-07-19
update nginx server name and README

jranks123:
Name changed from giraffe.thegulocal.com to contribute.thegulocal.com because it makes more sense to  keep the name in line with what it will be publicly. Also, updated README to instruct on what to add to their hosts file.


AWare:
👍  
Can you get a CSR for this and add the certs to the s3 bucket?

PR #11, opened 2016-07-21
Fix circular dep in Raven

AWare:
@desbo 

@tomverran / @paulbrown1982  this might want a look at in membership-frontend too


desbo:
👍 

I'll have a look at fixing the circular dep without needing the try/catch hack later on.

PR #13, opened 2016-07-21
remove facebook analytics (I'm not at all sure we use these)

AWare:
If we do, then there's more than just keeping these files needed to get it working.


jranks123:
DELETE 👍

PR #16, opened 2016-07-22
Thank you page

AWare:
- Adds the thank you page back
- Doesn't check the session cookie for donation
- Adds the social fragment
- Removes the duplicate contribute page from the thank you page template


rtyley:
:+1:

PR #17, opened 2016-07-25
Update amount variants

AWare:
Inline with current membership frontend


rtyley:
@aware assures me this is a direct copy and paste, what could go wrong!? :+1:

AWare:
nothing?

PR #19, opened 2016-07-25
removed duplicate variant from MessageCopyTest

pvighi:
I noticed there was a duplicate variant in MessageCopyTest


alexduf:
:+1:

PR #20, opened 2016-07-25
Variants by region

pvighi:
This PR adds the option of providing a list of country groups a variant applies to. When picking a variant on a specific request only variants that match the country group of the request will be considered. Variants defined with an empty country group filter will apply to all country groups.


app/views/support/ABTest.scala
```diff
@@ -90,27 +107,29 @@ case class ChosenVariants(v1: AmountHighlightTest.Variant, v2: MessageCopyTest.V
...
-  def pickByQueryStringOrCookie[A](request: Request[A], test: TestTrait): Option[test.Variant] = {
-    val search: Option[String] = request.getQueryString(test.slug)
-      .orElse(request.cookies.get(test.slug + "_GIRAFFE_TEST").map(_.value))
-    test.variants.list.find(_.variantSlug == search.getOrElse(None))
-  }
+  def pickVariant[A](countryGroup: CountryGroup, request: Request[A], test: TestTrait): test.Variant = {
+
+    def pickRandomly: test.Variant = {
+      val n = scala.util.Random.nextDouble()
+      test.weightedVariantsFor(countryGroup).dropWhile(_._1 < n).head._2
```
pvighi: This will crash horribly if we forget to include a variant that applies to the current countryGroup. 
If we want to avoid making this mistake we could either force the app to crash on startup if we end up with an empty list for any of the supported countryGroups or do headOption instead of head here and fall back to some default value

app/views/support/ABTest.scala
```diff
@@ -6,26 +6,51 @@ import java.net.URLEncoder
...
 
   def name: String
   def slug: String
   def variants: NonEmptyList[Variant]
-  val weight: Double = variants.map(_.weight).list.fold(0.0)(_ + _)
-  val cdf : Seq[Double]= variants.map(_.weight).list.foldLeft(Seq[Double]())( (l,p) => l :+ l.lastOption.getOrElse(0.0) + p/weight)
-  val weightedVariants = cdf.zip(variants.list)
+
+  def ValidateVariants = {
+    val countryWithNoVariants = variantsByCountry.collectFirst { case (country, variants) if variants.isEmpty => country.name }
```
alexduf: elegant :100:

app/views/support/ABTest.scala
```diff
@@ -90,27 +115,33 @@ case class ChosenVariants(v1: AmountHighlightTest.Variant, v2: MessageCopyTest.V
...
-    test.variants.list.find(_.variantSlug == search.getOrElse(None))
-  }
+  val allTests = List(AmountHighlightTest, MessageCopyTest)
+
+  def ValidateVariants = allTests.foreach(_.ValidateVariants)
+
+  def pickVariant[A](countryGroup: CountryGroup, request: Request[A], test: TestTrait): test.Variant = {
+
+    def pickRandomly: test.Variant = {
+      val n = scala.util.Random.nextDouble()
```
alexduf: while you're at it, could you add an `import scala.util.Random` ?

alexduf:
Looks good :+1: you probably want another pair of :eyes: on that: https://github.com/guardian/contributions-frontend/pull/20/files#diff-114818ea49160145b7f0d015ed576610R21

alexduf:
:+1:

app/views/support/ABTest.scala
```diff
@@ -6,26 +6,51 @@ import java.net.URLEncoder
...
-  val weightedVariants = cdf.zip(variants.list)
+
+  def ValidateVariants = {
+    val countryWithNoVariants = variantsByCountry.collectFirst { case (country, variants) if variants.isEmpty => country.name }
+    val errorMessage = countryWithNoVariants.map(country => s"${getClass.getName}: No variant defined for $country.")
+
+    errorMessage.foreach(m => throw new IllegalStateException(m))
+  }
+
+  val variantsByCountry: Map[CountryGroup, List[Variant]] = {
```
rtyley: The `List` type can probably be `Set`?

app/views/support/ABTest.scala
```diff
@@ -6,26 +6,51 @@ import java.net.URLEncoder
...
+
+    def filterVariants(countryGroup: CountryGroup) = variants.list.filter(_.matches(countryGroup))
+
+    CountryGroup.allGroups.map(country => country -> filterVariants(country)).toMap
+    }
+
+  val weightedVariantsByCountry = {
+
+    def addWeights(filteredVariants: List[Variant]) = {
```
rtyley: I like all this stuff! But because it's not that simple, maybe it's worth adding a couple of unit tests for this? Asserting that the cumulative distribution function adds up to 1, that sort of thing?

pvighi: Ok I'll add a unit test for this, and I will remove the runtime check with the ugly exception that I do in validateVariants and move it to this test as well

app/views/support/ABTest.scala
```diff
@@ -6,26 +6,51 @@ import java.net.URLEncoder
...
   def variants: NonEmptyList[Variant]
-  val weight: Double = variants.map(_.weight).list.fold(0.0)(_ + _)
-  val cdf : Seq[Double]= variants.map(_.weight).list.foldLeft(Seq[Double]())( (l,p) => l :+ l.lastOption.getOrElse(0.0) + p/weight)
-  val weightedVariants = cdf.zip(variants.list)
+
+  def ValidateVariants = {
+    val countryWithNoVariants = variantsByCountry.collectFirst { case (country, variants) if variants.isEmpty => country.name }
+    val errorMessage = countryWithNoVariants.map(country => s"${getClass.getName}: No variant defined for $country.")
+
+    errorMessage.foreach(m => throw new IllegalStateException(m))
```
markjamesbutler: Is there a better way of dealing with errors rather than using exceptions?  Could we enforce type safety so this fails at compile time rather than run time?

rtyley: > Is there a better way of dealing with errors rather than using exceptions? Could we enforce type safety so this fails at compile time rather than run time?

I think compile-time safety would be difficult, because the variants are defined with given countrygroups, and then you have to invert the map in order to find out which countries are not covered. Perhaps there is some kind of hardcore scalaz way of doing this, but I don't know what it is! However, it would be easy to create a unit test to ensure a bad setup fails at build time.

pvighi: Yeah I don't like this either.
 Since I'm already adding unit tests I'm adding one that checks that the generated variant weights make sense.

markjamesbutler: 👍  - Agreed, I'd definitely be interested in understanding how to implement these checks at compile time this in future.

app/views/support/ABTest.scala
```diff
@@ -38,7 +38,7 @@ trait TestTrait {
 
    def addRanges(filteredVariants: Set[Variant]): Seq[(Double, Variant)] = {
      val filteredVariantList = filteredVariants.toList
-     val weightSum: Double = filteredVariantList.map(_.weight).fold(0.0)(_ + _)
+     val weightSum: Double = filteredVariantList.map(_.weight).sum
```
alexduf: :+1: that looks nicer :)

rtyley:
:+1:

pvighi:
Seems to work fine ( but we have no country specific variants configured yet)

PR #22, opened 2016-07-26
Remove slashes from routes

jranks123:
The slashes were resulting in, for example, contribute.theguardian.com/uk not working, as it required contribute.theguardian.com/uk/

### DIFF
--- a/conf/routes
+++ b/conf/routes
@@ -11,11 +11,11 @@ GET            /assets/*file                controllers.Assets.versioned(path="/
 GET            /                            controllers.Giraffe.contributeRedirect
 GET            /healthcheck                 controllers.Healthcheck.healthcheck
 
-GET            /uk/                         controllers.Giraffe.contributeUK
-GET            /us/                         controllers.Giraffe.contributeUSA
-GET            /au/                         controllers.Giraffe.contributeAustralia
-GET            /eu/                         controllers.Giraffe.contributeEurope
-GET            /:countryGroup/              controllers.Giraffe.contribute(countryGroup: com.gu.i18n.CountryGroup)
+GET            /uk                         controllers.Giraffe.contributeUK
+GET            /us                         controllers.Giraffe.contributeUSA
+GET            /au                         controllers.Giraffe.contributeAustralia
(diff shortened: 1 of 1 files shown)


AWare:
This _is_ more in line with our other routes, but this is just exchanging one set of 404 errors for another.
👍

AWare:
This breaks the initial redirect.

jranks123:
ah. Sorry, must have missed that. Still on train, I'll fix it when I get home

conf/routes
```diff
@@ -11,11 +11,11 @@ GET            /assets/*file                controllers.Assets.versioned(path="/
...
-GET            /uk/                         controllers.Giraffe.contributeUK
-GET            /us/                         controllers.Giraffe.contributeUSA
-GET            /au/                         controllers.Giraffe.contributeAustralia
-GET            /eu/                         controllers.Giraffe.contributeEurope
-GET            /:countryGroup/              controllers.Giraffe.contribute(countryGroup: com.gu.i18n.CountryGroup)
+GET            /uk                         controllers.Giraffe.contributeUK
+GET            /us                         controllers.Giraffe.contributeUSA
+GET            /au                         controllers.Giraffe.contributeAustralia
+GET            /eu                         controllers.Giraffe.contributeEurope
+GET            /:countryGroup              controllers.Giraffe.contribute(countryGroup: com.gu.i18n.CountryGroup)
```
pvighi, line 19: This might be a stupid question but why do we need to explicitly set routes for uk, us au and eu if we have /:countryGroup? 
edit: Sorry I just realized this PR is already merged, my question was kinda off topic anyway for this PR

jranks123:
@pvighi Yes because sometimes people might want to explicitly visit a certain currency page despite not being in that location. For example, a UK citizen living in the United States

pvighi:
@jranks123 I don't think it makes a difference. after all you can switch to canada right now and there is no custom canada route

jranks123:
@pvighi  ah, yes you're right. I forgot about the currency switcher, it isn't in the old contributions page in membership-frontend

PR #23, opened 2016-07-26
update ga to new acc test + remove remarketing

AWare:


rtyley:
Deleting is always good! :+1: 

We should probably check that the `ANALYTICS_OFF_KEY` cookie still works though.

PR #24, opened 2016-07-26
Fix various problems with running on developer Ubuntu mach9ine

rtyley:
Various classpath and case-insensitive problems!

This isn't the config-refactoring @aware and I talked about - I'm going to do that next.


AWare:
:+1:

AWare:
Oops, not my branch to prune.

PR #27, opened 2016-07-26
CSS tweaks - small text col change and dots colour change

akemitakagi:
Changed the small text colours and dots selected / not selected colours.


AWare:
👍

PR #28, opened 2016-07-26
remove dead js

AWare:


AWare:
I'm pretty sure that karma isn't running tests right now, so we probably should look at that. 

_It's also broken in membership-frontend_

tomverran:
I'm fairly sure the tests in membership frontend do actually work but if they don't you can always look at what we did for subs to remove all traces of RequireJS

https://github.com/guardian/subscriptions-frontend/commit/713531d3878640d0d361d849e2bbf0ce4376daa7

desbo:
👍

PR #29, opened 2016-07-26
Remove omniture and uet tracking

AWare:
We're not going to be using omniture for much longer. 
And no-one knows what UET is. 🙀 

They're both gone.


desbo:
👍

PR #32, opened 2016-07-27
Contribute no cache

alexduf:
renames a package that wasn't following the scala norm
adds no-cache header for a page that generates varying results.

Long-term wise this will need to happen on the client side, but that will do.

cc @rtyley @pvighi


pvighi:
👍

alexduf:
This is shelved until @rtyley introduces the filter that checks every endpoint

PR #33, opened 2016-07-28
Add support for test users

rtyley:
When a user is signed-in to Identity, and they have a valid test-user id, they get given the 'test' mode of Stripe. Otherwise, on production, they get the 'Live' mode as usua (note that in DEV, the user will get 'test' mode either way).

Updated Stripe config:

This uses a slightly simpler format than Membership projects, because we don't have the DEV/UAT/PROD backend distinction here.

Public and secret keys in Stripe are rotated separately, so storing them separately (with the public key in source control and the secret key in S3) seems reasonable to me.

cc @AWare


start-frontend-debug.sh
```diff
@@ -1,2 +1,2 @@
 #!/bin/bash
-sbt -Djava.awt.headless=true -jvm-debug 9998 "project frontend" "devrun"
+sbt -Djava.awt.headless=true -jvm-debug 9998 "devrun"
```
AWare, line 2: Nice catch.

AWare:
👍

rtyley:
You can see live stripe keys on PROD:

```
$ curl --silent https://contribute.theguardian.com/uk/ | grep stripePublicKey
        stripePublicKey: 'pk_live_auSwLB4KBzbN3JOUVHvKMe6f',
```

...but if you hit the page as a [test user](https://membership.theguardian.com/test-users), you get this in the page - a test stripe key, as desired:

```
stripePublicKey: 'pk_test_35RZz9AAyqErQshL410RDZMs',
```

tomverran:
Is there any chance of the pushing of this switching logic into the app being reconsidered? Having switching logic like this in every app means you have to wrap any library you pull in into something that needs to know about HTTP, and this is our _fourth_ implementation of such a switcher.

We've been happily using the multi router in the promotions tool for a while and it seems to work okay

PR #36, opened 2016-07-29
Add a filter which requires that every response has cache headers

rtyley:
This is borrowed from Membership, with a few tweaks to use Java 8's time classes. Note that only HTTP 200, 301, 404 responses are cached - you generally don't want to cache 500 errors that correspond to server crashes for instance.

Thanks to @alexduf for pointing out that the documentation at:

https://www.playframework.com/documentation/2.5.x/ScalaHttpFilters

...is totally unhelpful if you're using compile-time DI. I've deleted much of the application.conf where that same confusion occurred.

cc @AWare


app/filters/CheckCacheHeadersFilter.scala
```diff
@@ -0,0 +1,26 @@
...
+import javax.inject.{Inject, Singleton}
+
+import akka.stream.Materializer
+import controllers.Cached.suitableForCaching
+import play.api.Logger
+import play.api.mvc._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+@Singleton
```
alexduf, line 12: I think you don't need the javax.inject._

alexduf:
Looks good to me, but you probably want @AWare to have another pair of :eyes: to it :+1:

AWare:
🚢

rtyley:
Looks good on PROD, I was able to successfully contribute make contributions of £2000 & £456 as a test user:

image

https://dashboard.stripe.com/test/payments/ch_18dzbeCbpG0cQtlbkPUyXx6m

PR #37, opened 2016-08-02
Add the default Play security headers filter

rtyley:
https://www.playframework.com/documentation/2.5.x/SecurityHeaders

I've disabled CSP for the time being, we use inline scripts and enabling a sensible CSP while they're present looks tricky?

Added headers are:

```
X-Frame-Options: DENY
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
X-Permitted-Cross-Domain-Policies: master-only
```

cc @AWare


AWare:
:+1:

rtyley:
Successfully did a test-user donation on PROD:

image

image

PR #38, opened 2016-08-02
Remove tailing `/` (slash) from urls

rtyley:
This will give us:

https://contribute.theguardian.com/uk

...rather than:

https://contribute.theguardian.com/uk/

Note that the redirect code (that redirects from https://contribute.theguardian.com/ to a country-specific page) has been fixed, as that killed a previous attempt to remove these bad slashes: https://github.com/guardian/contributions-frontend/pull/25 .

cc @AWare @jranks123

### DIFF
--- a/app/controllers/Giraffe.scala
+++ b/app/controllers/Giraffe.scala
@@ -84,10 +84,10 @@ class Giraffe(paymentServices: PaymentServices) extends Controller {
   val maxAmount: Option[Int] = 2000.some
 
 
-  def contributeRedirect = /*NoCache*/Action { implicit request =>
+  def contributeRedirect = NoCacheAction { implicit request =>
     val countryGroup = request.getFastlyCountry.getOrElse(CountryGroup.RestOfTheWorld)
-    val url = MakeGiraffeRedirectURL(request, countryGroup)
-    Redirect(url, SEE_OTHER)
+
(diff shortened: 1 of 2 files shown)


app/controllers/Giraffe.scala
```diff
@@ -84,10 +84,10 @@ class Giraffe(paymentServices: PaymentServices) extends Controller {
   val maxAmount: Option[Int] = 2000.some
 
 
-  def contributeRedirect = /*NoCache*/Action { implicit request =>
+  def contributeRedirect = NoCacheAction { implicit request =>
     val countryGroup = request.getFastlyCountry.getOrElse(CountryGroup.RestOfTheWorld)
-    val url = MakeGiraffeRedirectURL(request, countryGroup)
-    Redirect(url, SEE_OTHER)
+
+    Redirect(routes.Giraffe.contribute(countryGroup).absoluteURL(), SEE_OTHER)
```
AWare, line 90: 🔥on fire🔥

AWare:
👍

app/controllers/Giraffe.scala
rtyley, line 90: Ah, using `.absoluteURL()` means that the redirect goes to HTTP rather than HTTPS (because the Play app server thinks the request is insecure) - I'm going to change this to just `.url` with bba503b2dd62071087405eacae4e49ee0a3229a5.

PR #41, opened 2016-08-02
fixed Facebook and daily email links and remove duplicated code

pvighi:
Note that using data-shown this way we must add options for nzd and cad which is now allowed in contributions-frontend


rtyley:
:+1:

pvighi:
checked

PR #42, opened 2016-08-02
add australia to navigation

pvighi:
The country selector at the top of the page doesn't currently display Australia as an option.  If you go to the australian url you will see Australia selected but it will not be an option in the dropdown and you will not be able to get back if you change countries (unless you manually change the url)


AWare:
👍

pvighi:
checked

PR #43, opened 2016-08-02
Fix T&C's links

markjamesbutler:
The links are broken on PROD: 

https://contribute.theguardian.com/@Links.giraffeTerms

This is due to a small error in the template that was introduced in the refactor.  This PR fixes this issue.


desbo:
👍

PR #44, opened 2016-08-02
Fix redirect on / to include campaign codes

markjamesbutler:
The redirect from:

(https://contribute.theguardian.com/?INTCMP=article-1-everyone)

 Is missing campaign codes:

(https://contribute.theguardian.com/uk)

This PR ensures that the campaign codes are retained on the redirect.


rtyley:
:+1: :+1:

PR #45, opened 2016-08-03
Update webpack config to allow use of webpack-dev-server

desbo:
A couple of tweaks to the configs so the dev server works. This proxies all requests to the backend but allows for live reloading of files in the bundle. The command is:

```
webpack-dev-server --port 7777
```

You can then access `http://localhost:7777/webpack-dev-server/uk-react/` for hot reloading goodness! 

@AWare

### DIFF
--- a/webpack.config.prod.js
+++ b/webpack.config.prod.js
@@ -1,8 +1,9 @@
 var Uglify = require("webpack/lib/optimize/UglifyJsPlugin");
+var path = require('path');
 
-module.exports = function(debug) { return {
+module.exports = {
     resolve: {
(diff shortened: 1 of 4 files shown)


webpack.config.prod.js
```diff
@@ -51,6 +64,5 @@ module.exports = function(debug) { return {
     },
 
     context: 'assets/javascripts',
-    debug: false,
-    devtool: 'source-map'
```
AWare, line 55: Source maps are used by sentry

AWare:
👍

PR #46, opened 2016-08-03
Fix raven/sentry config

AWare:
- Use **correct** dsn
- Whitelist contributions

NEVERMORE


rtyley:
Looks good :+1: 

So all this goes into https://app.getsentry.com/the-guardian/contributions-js/ - right?

PR #47, opened 2016-08-03
Customisation for australia thank you page

pvighi:
and added nzd and cad to "data-shown" options

### DIFF
--- a/app/controllers/Giraffe.scala
+++ b/app/controllers/Giraffe.scala
@@ -117,12 +117,16 @@ class Giraffe(paymentServices: PaymentServices) extends Controller {
       .withCookies(Test.createCookie(chosenVariants.v1), Test.createCookie(chosenVariants.v2))
   }
 
-
   def thanks(countryGroup: CountryGroup) = NoCacheAction { implicit request =>
-    val redirectUrl = routes.Giraffe.contribute(countryGroup).url
 
+    val title = countryGroup match {
+      case CountryGroup.Australia => "Thank you for supporting Guardian Australia"
+      case _ => "Thank you for supporting the Guardian"
+    }
+
+    val redirectUrl = routes.Giraffe.contribute(countryGroup).url
     Ok(views.html.giraffe.thankyou(PageInfo(
-      title = "Thank you for supporting the Guardian",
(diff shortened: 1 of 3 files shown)


AWare:
👍

app/views/giraffe/thankyou.scala.html
```diff
@@ -10,9 +10,12 @@ <h1 class="support-thanks__title">@pageInfo.title</h1>
             <p class="support-thanks__message">@pageInfo.description</p>
 
             <h1 class="support-thanks__title">Spread the word</h1>
-            <p class="support-thanks__message">
+            <p class="support-thanks__message" data-shown="gbp eur usd nzd cad">
                 We report for everyone. Do tell your friends and family how they can support the Guardian too.
             </p>
+            <p class="support-thanks__message" data-shown="aud">
+                Your contribution is much appreciated, and will help us to maintain our fearless, independent, journalism in Australia.
+            </p>
```
rtyley, line 18: As you mentioned yesterday, it's a bit weird that this is done client-side! Oh well...

app/controllers/Giraffe.scala
```diff
@@ -117,12 +117,16 @@ class Giraffe(paymentServices: PaymentServices) extends Controller {
       .withCookies(Test.createCookie(chosenVariants.v1), Test.createCookie(chosenVariants.v2))
   }
 
-
   def thanks(countryGroup: CountryGroup) = NoCacheAction { implicit request =>
-    val redirectUrl = routes.Giraffe.contribute(countryGroup).url
 
+    val title = countryGroup match {
+      case CountryGroup.Australia => "Thank you for supporting Guardian Australia"
```
rtyley, line 123: We should keep an eye on the number of branches we have like this - might end up modelled as a case class for specific branding.

rtyley:
It would be great to remove the bizarre scss method for choosing thank you text, but if you want to make the smallest change :+1:

pvighi:
yeah I agree it's not ideal but the Australia changes need to be out as soon as possible so there's no time to refactor everything. So for now I'm making the smallest change

pvighi:
checked

PR #48, opened 2016-08-03
Use SAMEORIGIN for X-FRAME-OPTIONS

desbo:
This will allow us to use webpack-dev-server locally and still prevent anyone framing the page.
`ALLOW-FROM` is [badly supported](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options) so I think this is a better way. @rtyley


rtyley:
:+1: this matches https://github.com/guardian/subscriptions-frontend/blob/1ec90e36/app/Global.scala#L12

PR #52, opened 2016-08-03
add robots.txt

AWare:

PR #53, opened 2016-08-03
Fix robots.txt by providing a route to it...

rtyley:
Follow on from https://github.com/guardian/contributions-frontend/pull/52

cc @AWare

### DIFF
--- a/conf/routes
+++ b/conf/routes
@@ -2,6 +2,7 @@
 # This file defines all application routes (Higher priority routes first)
 # ~~~~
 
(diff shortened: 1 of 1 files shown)


AWare:
👍

rtyley:
Verified, this works: https://contribute.theguardian.com/robots.txt ...and other pages like https://contribute.theguardian.com/uk are still accessible.

We need to set up https://www.google.com/webmasters/tools/home to get a full overview.

PR #55, opened 2016-08-04
Add Google Webmasters tools validation

rtyley:
https://support.google.com/webmasters/answer/35179?authuser=0

The file must be kept in place in order for validation to persist.

cc @AWare


AWare:
:+1:

PR #56, opened 2016-08-04
remove max width for tablet

AWare:
Removing the max width fixed the weirdness Mario was seeing. 

image

image

@desbo @markjamesbutler


desbo:
👍

PR #58, opened 2016-08-05
ignore more of public

AWare:
I removed the public rule from gitignore now that 🤖.txt lives there
Maybe it would be more sustainable to put robots.txt in assets and have grunt copy it and just ignore the whole public dir.


rtyley:
:+1: I'm pretty sure that `public` is not actually meant to be a build directory.

AWare:
Blame javascript. :( 
_Merges_

PR #59, opened 2016-08-05
actually fix the tablet breakpoint

AWare:
- Deal with hidden-tablet and hidden-mobile being the same
- Deal with it some more
- Make the guardian circular button not shown for tablets
- Give the constrained column 100% until tablet to fix weirdness from flexbox


AWare:
ok I broke something with this 
update: fixed

pvighi:
👍

AWare:
Looks like it's working :)

PR #61, opened 2016-08-05
show some copy to canadians

AWare:
Currently we don't show anything to Canadians, lets show them what we show Americans.
before:
image
after:
image

before:
image
after:
image


markjamesbutler:
👍

pvighi:
looks good 👍 
by the way we have the same problem with nz, I don't know if we care about that

AWare:
@pvighi We'll have to fix that when we add nz to the dropdown. 
image

PR #63, opened 2016-08-08
Fixes to ga code

AWare:
- Fix a typo
- Make ophan stuff async
- Add some guards 

@ajosephides

### DIFF
--- a/assets/javascripts/src/modules/analytics/setup.js
+++ b/assets/javascripts/src/modules/analytics/setup.js
@@ -30,7 +30,6 @@ define([
 
     function setupAnalytics() {
         ophan.init();
-        ga.init();
         ophan.loaded.then(ga.init,ga.init);
     }
 
--- a/assets/javascripts/src/modules/analytics/setup.js
+++ b/assets/javascripts/src/modules/analytics/setup.js
@@ -31,9 +31,7 @@ define([
     function setupAnalytics() {
         ophan.init();
         ga.init();
-        ophan.ophan.then(function(ophan){
-            ga.setOphanId(ophan.pageViewId);
-        },null)
+        ophan.loaded.then(ga.init,ga.init);
     }
 
     function setupThirdParties() {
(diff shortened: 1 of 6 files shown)


AWare:
@paulbrown1982  @cb372

ajosephides:
Fine with this to go with the other bug fixes but with the Ophan async fix this will double the number of server calls to Google Analytics.
This is something we must avoid as it will have a cost implication and so it may be be better if we can have this loaded synchronously.

assets/javascripts/src/modules/analytics/setup.js
```diff
@@ -31,6 +31,9 @@ define([
     function setupAnalytics() {
         ophan.init();
         ga.init();
+        ophan.ophan.then(function(ophan){
+            ga.setOphanId(ophan.pageViewId);
```
paulbrown1982: Can't you just fire ga.init() here? If you only call 'setOphanId' here then it gets set after the pageview and won't be sent to Google. You'd have to fire pageview twice to make it send, which isn't appropriate.

AWare:
@paulbrown1982

assets/javascripts/src/modules/analytics/setup.js
```diff
@@ -31,6 +31,7 @@ define([
     function setupAnalytics() {
         ophan.init();
         ga.init();
+        ophan.loaded.then(ga.init,ga.init);
```
paulbrown1982: I'm just testing this, fwict it's firing both. Also you need to remove the ga.init() above!

assets/javascripts/src/modules/analytics/setup.js
```diff
@@ -31,6 +31,7 @@ define([
     function setupAnalytics() {
         ophan.init();
         ga.init();
```
paulbrown1982: You need to remove this one!

paulbrown1982:
👍

AWare:
I think my renaming ophan.ophan to ophan.loaded has broken things. Will have another look in the morning.

AWare:
That's better.
@paulbrown1982 ?

paulbrown1982:
👍

PR #64, opened 2016-08-08
Remove Canada from drop down. Redirect /ca and /nz to /uk

pvighi:
Get rid of Canada in drop down until cleared. And point Canada and New Zealand to UK page

### DIFF
--- a/app/controllers/Giraffe.scala
+++ b/app/controllers/Giraffe.scala
@@ -92,17 +92,20 @@ class Giraffe(paymentServices: PaymentServices) extends Controller {
   def contributeRedirect = NoCacheAction { implicit request =>
 
     val countryGroup = request.getFastlyCountry match {
-      case Some(Canada) | Some(NewZealand) => UK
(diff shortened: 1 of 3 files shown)


app/controllers/Giraffe.scala
```diff
@@ -84,20 +85,25 @@ class Giraffe(paymentServices: PaymentServices) extends Controller {
...
 
   def contributeRedirect = NoCacheAction { implicit request =>
-    val countryGroup = request.getFastlyCountry.getOrElse(CountryGroup.RestOfTheWorld)
+
+    val countryGroup = request.getFastlyCountry match {
+      case Some(Canada) | Some(NewZealand) => UK
+      case Some(other) => other
+      case None => RestOfTheWorld
+    }
+
```
pvighi, line 99: This change is not strictly necessary, it's just here to avoid redirecting twice. 
We could leave it as it is and it would redirect to /ca or /nz which in turn would redirect to /uk.

AWare:
Maybe we should also redirect int to uk?

alexduf:
:+1:

PR #66, opened 2016-08-09
The maximum donation amount is not optional, no matter what currency

rtyley:
Getting rid of a bunch of `folds` and Option-wrappers.

cc @AWare @pvighi

### DIFF
--- a/app/controllers/Giraffe.scala
+++ b/app/controllers/Giraffe.scala
@@ -1,32 +1,28 @@
 package controllers
 
+import java.lang.Math.min
 import java.net.URLDecoder
 import java.nio.charset.StandardCharsets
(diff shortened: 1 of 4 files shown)


AWare:
👍

rtyley:
@AWare remarks that we made this an option because of Patrons night, but if we were to want to do this again, I think we'd do it differently :smile:

rtyley:
Was able to contribute £2000 as a test-user on PROD:

https://dashboard.stripe.com/test/payments/ch_18gdVsCbpG0cQtlbOkMNJVgr

...the UI won't allow me to progress if I enter £2001, so that's good, though "We are presently only able to accept contributions of £2000 or less" is not highlighted in any way.

PR #68, opened 2016-08-10
Sentry release

AWare:
Sentry wants the build number in the release field. 
We're already sending this as a tag, which would now be superfluous. 

This enables release tracking in sentry. 
@mario-galic: you might want this for s-f and m-f


mariogalic:
👍  Nice! I'll enable it on our repos.

jranks123:
👍

PR #69, opened 2016-08-11
Changed cookie length from 2 weeks to 1 week

jranks123:
#304 in Trello.

The cookie expiration was 2 weeks. This meant that once someone was opted into an AB test, they couldn't be put into a new variant for a whole 2 weeks. This PR shortens that time to 1 week


alexduf:
:+1:

PR #70, opened 2016-08-16
Update return url for sign out

jranks123:
Previously, sign out was sending users to membership frontend rather than back to contribute. This fixes that issue.


markjamesbutler:
👍

PR #72, opened 2016-08-17
[WIP] Client-side React refactor 

desbo:
Sorry about the size of this PR! Here's what it contains (very broadly): 
- move the rendering of the contributions form from the server to client (using React/Redux)
- remove CSS/JS that we no longer need thanks to the above (there's probably more to remove)
- incorporate feedback from design and UX (including HTML5 validation and navigation fixes)

There are still a few things left to do:
- [x] fix Ophan module/get rid of curl
- [x] fix T&Cs styling
- [x] preselect amount based on query param (for some work @jranks123 is doing)
- [ ] device testing
- [x] add signed in user detection
- [ ] explain how it works to everyone

It's currently set to run on `/uk-react`. Although routes exist for the non-React (current) version of the app, the styles are broken due to the cleanup, so we should try to get this tested and merged to replace the existing version.


alexduf:
Excellent work @desbo ! that looks really good. :+1:

app/views/support/ABTest.scala
```diff
@@ -8,25 +8,54 @@ import java.nio.charset.StandardCharsets
...
+    }
   }
+}
 
+case class Variant(testName: String, testSlug: String, variantName: String, variantSlug: String, weight: Double, data: VariantData, countryGroupFilter: Set[CountryGroup] = Set.empty) {
+  def matches(countryGroup: CountryGroup): Boolean = countryGroupFilter.isEmpty || countryGroupFilter.contains(countryGroup)
+}
+
+object Variant {
+  implicit val writes: Writes[Variant] = (
```
alexduf, line 40: I just realised you can use the macro to do this one:

`implicit val writes = Json.writes[Variant]`

And that should work

app/views/support/ABTest.scala
```diff
@@ -8,25 +8,54 @@ import java.nio.charset.StandardCharsets
...
-    def testSlug = slug
-    def matches(countryGroup:CountryGroup):Boolean = countryGroupFilter.isEmpty || countryGroupFilter.contains(countryGroup)
+import views.support.AmountHighlightTest.AmountVariantData
+import views.support.CountryGroupImplicits._
+import views.support.MessageCopyTest.CopyVariantData
+
+sealed trait VariantData
+
+object VariantData {
+  implicit val formatter: Writes[VariantData] = new Writes[VariantData] {
```
alexduf, line 25: I think that variable should be renamed to `writes`

PR #73, opened 2016-08-17
[WIP] Paypal Integration

pvighi:
This PR adds a new Payment Methods AB test that determines if the user sees a PayPal payment button in addition to the regular card payment flow.
Client side code will probably we redone to work with the new react version of the site so not much effort was put into it (but should still work).

This branch has the PayPal AB test disabled so the button should never be visible but there are some differences:
1. The initial "next" button now says "contribute with card" and is grey instead of green
2. On mobile the payment details form is hidden initially and the "contribute with card" is shown instead, which must be clicked to reveal the form. This behaviour is not desired unless the PayPal option is enabled


app/controllers/PaymentError.scala
```diff
@@ -0,0 +1,15 @@
...
+}
+
+object PaymentError {
+  val allErrors: Map[String, PaymentError] = Map(PaypalError.toString -> PaypalError)
+
+  def fromString(errorCode: String): Option[PaymentError] = allErrors.get(errorCode)
+}
+
+case object PaypalError extends PaymentError {
+  override def message = "Your contribution using PayPal could not be processed. Please try again or use pay by card."
```
alexduf, line 14: Your contribution using PayPal could not be processed. Please try again or pay by card.

pvighi, line 14: I thought "use pay by card" was strange as well, but that's the text I've been given. 
I would have written "use another payment method" but maybe that's not specific enough. Even if we keep this line it might me more correct to say "use 'contribute with card'" to match the label in the button

app/controllers/PaypalController.scala
```diff
@@ -0,0 +1,105 @@
...
+      "countryGroup" -> of[CountryGroup],
+      "amount" -> bigDecimal(10, 2)
+    )(PaymentData.apply)(PaymentData.unapply)
+  )
+
+  def authorize = NoCacheAction { implicit request =>
+    paypalForm.bindFromRequest().fold[Result](
+      hasErrors = form => handleError(CountryGroup.UK, form.errors.mkString(",")),
+      success = form => {
+        val transactionId = transactionUtil.newTransactionId
```
alexduf: Do you mind if we call it contributionId? that's how I called it in the database.

app/controllers/PaypalController.scala
```diff
@@ -0,0 +1,105 @@
...
+        val paypalService = paymentServices.paypalServiceFor(request)
+        val maxAllowedAmount = configuration.Payment.maxAmountFor(form.countryGroup.currency)
```
alexduf, line 81: `import configuration.Payment`

app/utils/TransactionUtil.scala
```diff
@@ -0,0 +1,10 @@
+package utils
+import java.util.UUID
+
+trait TransactionUtil {
+  def newTransactionId: String
```
alexduf: Should this attribute keep the type UUID?

app/controllers/PaymentError.scala
alexduf, line 14: ah, fair enough then!

app/controllers/PaypalController.scala
```diff
@@ -0,0 +1,103 @@
...
+import play.api.Logger
+import play.api.data.format.Formatter
+import play.api.libs.json._
+import utils.ContributionIdGenerator
+import play.api.libs.functional.syntax._
+
+import scala.util.Right
+
+
+class PaypalController(ws: WSClient, paymentServices: PaymentServices, contributionIdGenerator :ContributionIdGenerator) extends Controller {
```
alexduf: I'm being pedantic but the spacing for the parameter contributionIdGenerator is wrong

PR #77, opened 2016-08-23
Second page copy logic for contributions interactive embed

jranks123:
When the user comes in from the contributions interactive embed (CIB), they have already chosen their amount so they need to be taken to the second page of the form and see different copy acknowledging that they have already chosen their contribution amount, unless they are on mobile, where this is not applicable

This PR implements this copy logic. 

screenshot at aug 23 13-31-47

^ What the second page looks like if you come from the CIB on tablet or larger

screenshot at aug 23 13-32-41

^ What the second page looks like if you come from a normal link

screenshot at aug 23 13-32-28

^ What the form looks like on mobile, regardless of where you came from

### DIFF
--- a/app/views/giraffe/contribute.scala.html
+++ b/app/views/giraffe/contribute.scala.html
@@ -69,9 +69,9 @@
                     </div>
 
                     <div class="form__column js-hidden-tablet js-details">
-                        <h3 class="form__column-heading form__column-heading--first headline-text--bolder details-heading doNotSkip">
+                        <h3 class="form__column-heading form__column-heading--first headline-text--bolder details-heading js-default-heading">
                             Your details</h3>
-                        <h3 class="form__column-heading form__column-heading--first headline-text--bolder details-heading doSkip" >
+                        <h3 class="form__column-heading form__column-heading--first headline-text--bolder details-heading js-amount-heading" >
                             Continue with your
                                 <br>
                                 contribution of <span class="js-currency-pay">@countryGroup.currency.prefix@countryGroup.currency.glyph</span><span class="js-amount js-pay"></span>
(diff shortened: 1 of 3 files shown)


app/views/giraffe/contribute.scala.html
```diff
@@ -69,8 +69,14 @@ <h3 class="form__column-heading headline-text--bolder">Your contribution</h3>
                     </div>
 
                     <div class="form__column js-hidden-tablet js-details">
-                        <h3 class="form__column-heading form__column-heading--first headline-text--bolder">
+                        <h3 class="form__column-heading form__column-heading--first headline-text--bolder details-heading doNotSkip">
```
AWare: doNotSkip could be the default state, so does not need to present here

assets/stylesheets/giraffe.scss
```diff
@@ -252,6 +253,32 @@ $desktop-col-padding-right: 16px;
...
+    .doNotSkip.js-skip {
+        display:none;
+    }
+}
+
+
+@include mq($until: tablet) {
+    .doSkip.js-skip {
+        display:none;
+    }
```
AWare, line 274: js- hyphen classes shouldn't have styles associated with them 

edit: just rename these classes

app/views/giraffe/contribute.scala.html
```diff
@@ -69,8 +69,14 @@ <h3 class="form__column-heading headline-text--bolder">Your contribution</h3>
                     </div>
 
                     <div class="form__column js-hidden-tablet js-details">
```
AWare, line 71: I should not have used js- for a class I was styling off. :(

AWare:
There's some misuse of js- classes with css, which I'm completely to blame for starting. As we're going to merge react soon, I don't think it's worth refactoring the class names.
:+1:

PR #78, opened 2016-08-24
Paypal meta data

alexduf:
This change is saving meta data and the contributor identity when a transaction has been executed.

cc @pvighi

### DIFF
--- a/app/services/PaypalService.scala
+++ b/app/services/PaypalService.scala
@@ -100,7 +100,6 @@ class PaypalService(config: PaypalApiConfig, contributionData: ContributionData)
       if (createdPayment.getState.toUpperCase != "APPROVED") {
         Left(s"payment returned with state: ${createdPayment.getState}")
       } else {
-        Payment.get(apiContext, paymentId)
         Right(createdPayment.getId)
       }
     } catch {
--- a/app/services/PaypalService.scala
+++ b/app/services/PaypalService.scala
@@ -1,5 +1,7 @@
 package services
 
+import java.util.UUID
+
 import com.gu.i18n.CountryGroup
 import com.paypal.api.payments._
 import com.paypal.base.Constants
@@ -7,6 +9,13 @@ import com.paypal.base.rest.{APIContext, PayPalRESTException}
 
 import scala.collection.JavaConverters._
 import com.typesafe.config.Config
+import data.ContributionData
+import models.ContributionMetaData
+import org.joda.time.DateTime
+import play.api.Logger
+import views.support.ChosenVariants
+
+import scala.util.{Failure, Success, Try}
 
 case class PaypalCredentials(clientId: String, clientSecret: String)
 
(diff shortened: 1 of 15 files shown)


app/services/PaypalService.scala
```diff
@@ -59,27 +87,69 @@ class PaypalService(config: PaypalApiConfig) {
...
-  def executePayment(paymentId: String, token: String, payerId: String): Either[String, Unit] = {
+  def executePayment(paymentId: String, payerId: String): Either[String, String] = {
     val payment = new Payment().setId(paymentId)
     val paymentExecution = new PaymentExecution().setPayerId(payerId)
     try {
       val createdPayment = payment.execute(apiContext, paymentExecution)
       if (createdPayment.getState.toUpperCase != "APPROVED") {
         Left(s"payment returned with state: ${createdPayment.getState}")
+      } else {
+        Payment.get(apiContext, paymentId)
```
pvighi: why do you call Payment.get here ?

app/services/PaypalService.scala
```diff
@@ -59,27 +87,69 @@ class PaypalService(config: PaypalApiConfig) {
...
+        Right(createdPayment.getId)
       }
-      else
-        Right(Unit)
     } catch {
       case e: PayPalRESTException => Left(e.getMessage)
     }
   }
 
+  def storeMetaData(
```
pvighi, line 110: Is this specific to PayPal  or could be have it in a more general place where we could use it for stripe in the future?

app/services/PaypalService.scala
alexduf: woups :D

app/services/PaypalService.scala
alexduf: I think it's completely specific. The best we could would be to extract it to a trait, but even then, I doubt the signature would match with what stripe has

app/services/PaypalService.scala
```diff
@@ -59,27 +87,69 @@ class PaypalService(config: PaypalApiConfig) {
...
       approvalLink.map(l => Right(l.getHref)).getOrElse(Left("No approval link returned from paypal"))
-    }
-    catch {
+    } catch {
       case e: PayPalRESTException => Left(e.getMessage)
     }
   }
 
-  def executePayment(paymentId: String, token: String, payerId: String): Either[String, Unit] = {
+  def executePayment(paymentId: String, payerId: String): Either[String, String] = {
```
pvighi, line 95: Does this now return a payment id on success just because it feels wrong to return unit?

alexduf: Yeah I was bugged by that `Either[String, Unit]` even though I doubt it's actually wrong

app/services/PaypalService.scala
pvighi: ok

pvighi:
👍

PR #79, opened 2016-08-24
always display the "secure" label and the padlock

alexduf:
Changed the copy to "Secure" and always display it.

cc @desbo


desbo:
👍

PR #80, opened 2016-08-24
add "(optional)" to the postcode field

alexduf:
Looks like that
screenshot from 2016-08-24 16-07-34

Please @desbo let me know if I'm implementing it in a sensible way

### DIFF
--- a/assets/javascripts/src/components/InputField.jsx
+++ b/assets/javascripts/src/components/InputField.jsx
@@ -2,11 +2,11 @@ import React from 'react';
 
 export default class InputField extends React.Component {
     render() {
-        const { outerClassName, children, optional, ...props } = this.props;
+        const { outerClassName, children, ...props } = this.props;
 
         return <div className={outerClassName || ''}>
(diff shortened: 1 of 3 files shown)


assets/javascripts/src/components/InputField.jsx
```diff
@@ -2,10 +2,12 @@ import React from 'react';
 
 export default class InputField extends React.Component {
     render() {
-        const { outerClassName, children, ...props } = this.props;
+        const { outerClassName, children, optional, ...props } = this.props;
 
         return <div className={outerClassName || ''}>
-            <label htmlFor={this.props.id} className="label">{this.props.label}</label>
+            <label htmlFor={this.props.id} className="label">{this.props.label}
+                {optional && <span className="label-optional">(optional)</span>}
```
desbo: There's no need to pass `optional` as an extra prop, you can check against the standard `required` attribute which is passed to `InputField` as `this.props.required`

desbo:
👍

PR #85, opened 2016-08-31
Paypal react

pvighi:
If you update your local contributions.private.conf you should be able to run this locally.
The only thing that will not work is the redirection back to contributions frontend after a paypal payment. To make this work just change the variable baseReturnUrl in the configuration to an externally accessible url that points to your local machine. I use [ngrok](https://ngrok.com/) to get a url that points to my local instance of contributions frontend.

Main PayPal flow:
1. A contributor clicks on the PayPal button
2. An ajax request is posted to [contributions-frontend]/paypal/auth specifying the amount and countryGroup
3. The approvalUrl returned in the response is used to redirect the contributor to PayPal to authorize the payment
4. After the payment is authorized paypal will redirect the user to [contributions-frontend]//paypal/[countryGroup]/execute passing payment & payer ids as parameters
5. A PayPal API call is executed to finally execute the payment passing payer and payment ids, if everything goes well metadata about the contribution is stored in the db (but not the result of the payment) and the user is redirected to the thank you page
6. When the payment is finalized a webhook is triggered in paypal that will POST to [contributions-frontend]/paypal/hook. This will store additional data about the payment result


app/views/giraffe/contribute.scala.html
```diff
@@ -27,10 +30,19 @@ <h2 class="giraffe__heading--sub">
...
+                        @message
+                        <a class="action action--contribute js-advance" id="error_tryAgain" >Try again</a>
+                </div >
+            }
+
+            <div class="form__container @containerVisibility">
                 <div class="form__switcher">
                     <a href="#" class="js-switcher-contribute form__container--active" data-switches="js-contribution"></a> <a class="js-switcher-details" href="#" data-switches="js-details"></a> <a class="js-switcher-pay" href="#" data-switches="js-payment"></a>
                 </div>
+                @fragments.giraffe.paypalForm(countryGroup, cmpCode, intCmpCode)
```
desbo: I guess we don't need any of this stuff  in this file or any of the fragments now?

pvighi: You are right, I forgot to remove all the fragments that are no longer in use.
Do we even need to keep  contribute.scala.html ? I just removed the code that didn't compile after the merge but other than that I ignored it

assets/javascripts/src/actions.es6
```diff
@@ -39,6 +43,32 @@ export function submitPayment(dispatch) {
         .catch(error => dispatch({ type: PAYMENT_ERROR, kind: 'network', error: error }));
 }
 
+export function paypalRedirect(dispatch) {
+    let state = store.getState();
```
desbo: this isn't a blocker for merging, but it's better to use `const` here instead, only use `let` if you want to reassign the variable (`const` == `val`, `let` == `var`)

assets/javascripts/src/actions.es6
```diff
@@ -39,6 +43,32 @@ export function submitPayment(dispatch) {
...
+
+    dispatch({ type: SUBMIT_PAYMENT });
+
+    let postData = {
+            countryGroup: state.data.countryGroup.id ,
+            amount: state.card.amount //TODO should the amount be somewhere else rather than in the card section?
```
desbo, line 53: it probably should!
maybe in data

app/views/giraffe/contribute.scala.html
desbo: Nope, there's no need for any of it, we can just rename `contributeReact` to `contribute`

assets/stylesheets/giraffe-react.scss
```diff
@@ -146,9 +221,9 @@
 }
 
 .contribute-navigation__next {
-    @include mq($until: tablet) {
-        display: none;
-    }
+  //  @include mq($until: tablet) {
+  //      display: none;
+  //  }
```
desbo: 👀 
do we need these lines anymore?

app/data/ContributionData.scala
```diff
@@ -0,0 +1,114 @@
...
+          ophanid = excluded.ophanid,
+          abtests = excluded.abtests,
+          cmp = excluded.cmp,
+          intcmp = excluded.intcmp"""
+      request.execute()
+    }
+  }
+
+  def insertContributor(contributor: Contributor): Unit = {
+    db.withConnection(autocommit = true) { implicit conn =>
```
AWare, line 83: I'm not keen on us directly accessing an SQL database from contributions. I'd be happier if we had some kind of API endpoint or queue instead. Also, directly including the SQL logic here might make things harder when we want a secondary data store to handle end user requests. (ie history/contributions status)

alexduf: I understand the API argument, in fact I started using api gateway / lambdas but stopped midway as it was getting fiddly and painful to use. I took a more pragmatic approach as we're running out of time and double checked with @markjamesbutler

It is not excluded to move that code into an API behind a different domain name, I too like separation of responsibility. As a side note It wouldn't bring more in terms of security as the API would be exposed on the Internet anyway (web hooks) but it would be more elegant to put backend considerations away from the frontend. 

I'm not sure about the queue argument though. I don't think that would answer any of our needs.

app/services/PaypalService.scala
```diff
@@ -0,0 +1,157 @@
...
+    transaction.setCustom(contributionId)
+    transaction.setItemList(itemList)
+
+    val transactions = List(transaction).asJava
+
+    val payer = new Payer().setPaymentMethod("paypal")
+    val redirectUrls = new RedirectUrls().setCancelUrl(cancelUrl).setReturnUrl(returnUrl)
+
+    val payment = new Payment().setIntent("sale").setPayer(payer).setTransactions(transactions).setRedirectUrls(redirectUrls)
+    try {
```
markjamesbutler: Using the Try class here will allow the compile to type check the code rather than the key word try.  

[Error handling with Try](http://danielwestheide.com/blog/2012/12/26/the-neophytes-guide-to-scala-part-6-error-handling-with-try.html)

app/utils/ContributionIdGenerator.scala
```diff
@@ -0,0 +1,10 @@
+package utils
+import java.util.UUID
+
+trait ContributionIdGenerator {
+  def getNewId: String
+}
+
+object ContributionIdGeneratorImpl extends ContributionIdGenerator {
```
markjamesbutler: It looks like the trait isn't used here.  Unless you are using this for testing (which I don't think you are) it might be worth simply calling the object rather than injecting it into the controller then removing the trait.

app/views/support/ABTest.scala
```diff
@@ -139,6 +140,25 @@ object MessageCopyTest extends TestTrait {
...
+  def slug = "paymentMethods"
+
+  case class PaymentMethodVariantData(paymentMethods: Set[String]) extends VariantData
+
+  object PaymentMethodVariantData {
+    implicit val format: Format[PaymentMethodVariantData] = Json.format[PaymentMethodVariantData]
+  }
+
+  def variants = NonEmptyList(
+    makeVariant("Control", "control", 1, PaymentMethodVariantData(Set("CARD"))),
```
markjamesbutler: paymentMethods could be a trait with Card and PayPal as case objects to give compile time safety instead of using Strings.  Obviously you'd need to fix the JS code if you made this change.

conf/DEV.public.conf
```diff
@@ -17,7 +17,14 @@ stripe {
         }
     }
 }
-
+paypal {
+    default=TEST
```
AWare, line 21: I'm not sure that paypal should be split like this instead of using the existing DEV->UAT->PROD split config. (With DEV & UAT -> TEST)

Edit: OK this is how we do this now, must have missed it.

build.sbt
```diff
@@ -27,26 +27,32 @@ sources in (Compile,doc) := Seq.empty
...
     scalaUri,
     scalaz,
     membershipCommon,
     sentryRavenLogback,
     memsubCommonPlayAuth,
+    paypalSdk,
     "com.softwaremill.macwire" %% "macros" % "2.2.2" % "provided",
     "com.softwaremill.macwire" %% "util" % "2.2.2",
-    "com.softwaremill.macwire" %% "proxy" % "2.2.2"
+    "com.softwaremill.macwire" %% "proxy" % "2.2.2",
```
AWare: Maybe we should name all of our imports?

alexduf:
Excellent work !  :+1: 
giphy

PR #87, opened 2016-08-31
Remove compile_js from grunt:watch task

desbo:
We don't need this because `webpack-dev-server` takes care of compiling JS in dev (via `npm run watch`).


alexduf:
I'll trust you on this change as I have no clue how these things work :D :+1:

PR #90, opened 2016-08-31
Fix A/B test preset amount overriding the one from query params

desbo:
Currently the A/B test to highlight an amount overrides the one set in the query string. This fixes that.


alexduf:
:+1:

PR #93, opened 2016-09-01
turn marketing on by default

alexduf:


alexduf:
cc @desbo

desbo:
👍

desbo:
although calling it `optIn` may be a little inaccurate now...

PR #94, opened 2016-09-02
fix form submission ignoring input value

alexduf:
this should fix the case where a user sets a value, presses return, only to see its value ignored as the onBlur event has not been fired.

cc @desbo

### DIFF
--- a/assets/javascripts/src/components/pages/Contribution.jsx
+++ b/assets/javascripts/src/components/pages/Contribution.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 
 export default class Contribution extends React.Component {
     constructor(props) {
-        super(props)
+        super(props);
 
         this.state = {
             highlightButton: true,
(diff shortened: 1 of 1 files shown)


assets/javascripts/src/components/pages/Contribution.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 
 export default class Contribution extends React.Component {
     constructor(props) {
-        super(props)
```
markjamesbutler, line 5: Are we running JSLint etc to automatically catch this type of error?

markjamesbutler:
👍  LGTM

assets/javascripts/src/components/pages/Contribution.jsx
alexduf: To be honest I'm not sure. The js build pipeline is still dark magic to me :smile:

PR #95, opened 2016-09-06
Test payments not working

pvighi:
On the react version of the frontend we execute payments as ajax requests using fetch.
Fetch will not send cookies in a post unless a specific option is provided. 
This probably makes our posts unauthenticated so we get the live version of all services.
https://github.com/github/fetch#sending-cookies


alexduf:
Cool! let's test that :+1:

pvighi:
Looks like it works. 
it uses test PayPal and Stripe if logged in as a test user

PR #97, opened 2016-09-07
Allow decimals in custom contribution amount

desbo:
You can now enter decimals in the contribution field!! 🎊

### DIFF
--- a/assets/javascripts/src/components/pages/Contribution.jsx
+++ b/assets/javascripts/src/components/pages/Contribution.jsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import { formatCurrency } from 'src/utils/formatters';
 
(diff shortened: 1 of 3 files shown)


pvighi:
Entering amounts with more than 2 decimal places causes an error when calling the PayPal API:

> "Currency amount must be non-negative number, may optionally contain exactly 2 decimal places separated by '.', optional thousands separator ',', limited to 7 digits before the decimal point and currency which is a valid ISO Currency Code"

If you are not in a hurry I can fix the PayPal service to format the amount correctly before making the call. But we should not merge this before that is fixed

desbo:
Thanks @pvighi, I've changed the code to do the currency formatting in the reducer rather than in the components.

alexduf:
:+1:

PR #99, opened 2016-09-07
terms & conditions and privacy policy

pvighi:
- Changed format
- fixed australia link
- added links to t&c and privacy policy in the disclamer section so PayPal users can see it

legalbefore
legal


pvighi:
We still need to fix the fact that changing countries from the dropdown doesn't update the non-react parts of the contribution page so the terms and conditions link will not be updated.

alexduf:
I think we should fix the problem you are mentioning by using react at the bottom of the page as well. (but that's for another PR)

app/views/fragments/giraffe/legalNotice.scala.html
```diff
@@ -0,0 +1,4 @@
+@(termsAndConditionsLink: String)()
+    <p>By proceeding, you are agreeing to our <a href="@termsAndConditionsLink"> Terms & Conditions</a>
```
alexduf: `&` => `&amp;`

AWare:
:+1:

alexduf:
:+1:

PR #100, opened 2016-09-08
Add copy:curl to runWatch task :flushed:

desbo:
I somehow ended up removing the one thing that my last change was meant to introduce. 
This fixes it. 
@alexduf @pvighi


alexduf:
@AWare will be happy to see an emoji in the commit :smile:

alexduf:
:+1:

PR #101, opened 2016-09-08
Update URL on region change

desbo:
Use `history.replaceState` to change the URL when switching region. Also move a few functions from `contribute.es6` to their own module.

sep 08 2016 11 32


markjamesbutler:
LGTM 👍

PR #105, opened 2016-09-08
quick fix for validation bug

pvighi:
we should rethink how we trigger paypal/stripe payments. This is a quick fix that touches the least amount of code in order to minimize risk


desbo:
👍

PR #106, opened 2016-09-08
IE fixes

desbo:
Some stuff to make it work in IE:
1. Add `Array.prototype.find` polyfill
2. Use `getAttribute` rather than (badly supported) `dataset` 
3. Add babel es2015 class transform


markjamesbutler:
LGTM 👍

PR #109, opened 2016-09-08
Fix promise polyfill

desbo:
The previous approach didn't work.


markjamesbutler:
👍 Shame promise-polyfill didn't work.

desbo:
Yup, this is actually the same library but I guess importing it in that way doesn't work :/

PR #115, opened 2016-09-09
push Paypal test to 50/50

alexduf:
Data is looking good, so let's push it to the 50/50 value in preparation for the week-end


alexduf:
cc @pvighi

pvighi:
👍

PR #121, opened 2016-09-13
Update and fix webpack-dev-server

desbo:
The syntax for the proxy config [has changed](https://github.com/webpack/webpack-dev-server/issues/562) in the current version of webpack-dev-server.
Although we specify the version in package.json, this actually needs to be installed globally (ignoring package.json version), so it's better to make it compatible with the latest one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "time-grunt": "^1.4.0",
     "vinyl-fs": "2.4.3",
     "webpack": "^1.13.1",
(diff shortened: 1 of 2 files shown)


AWare:
:+1: 

There are some slight issues regarding running the backend without https. 
https://giraffe.thegulocal.com points at webpack dev server. (But could do with a proper cert)

AWare:
Oh, also, we might want to remove the ^ from the version string and avoid this problem happening next time.

PR #125, opened 2016-09-15
Ga dropoff tracking

alexduf:
- Each pages has its own javascript
- Refactor the event tracking to support promises and therefore order events
- Split analytics initialisation and pageview firing
- Register form events to track drop-off

cc @desbo Please don't hesitate to be critical on that PR, my react knowledge is very fresh :smiley:


assets/javascripts/src/modules/analytics/ga.es6
```diff
@@ -16,8 +16,10 @@ const dimensions = {
     intcmp: 'dimension12'
 };
 
+const defaultTracker = 'membershipPropertyTracker';
+var gaEnabled = true;
```
alexduf: This bit I'm not sure about... I need to somehow hold a state for browsers or users that don't want to be tracked without breaking later when we call the ga function (via gaProxy)

assets/javascripts/src/modules/analytics/ga.es6
```diff
@@ -33,13 +35,25 @@ function create(){
...
         'cookieDomain': guardian.googleAnalytics.cookieDomain
     });
-    return (a,b,c) => window.ga(tracker+ '.' + a,b,c);
+    return gaProxy
 }
 
-export function init() {
+export function gaProxy() {
+    if (gaEnabled) {
+        let allArgs = Array.from(arguments);
```
desbo: just FYI really but you can still use `const` here even though you're mutating the array - arrays (and objects) are always mutable in JS, `const` just means that you can't reassign it, whereas you can with `let`.  it's always better to use `const` unless you're reassigning the variable later on.

assets/javascripts/src/components/Main.jsx
```diff
@@ -48,12 +49,24 @@ function mapDispatchToProps(dispatch) {
     return {
         goBack: () => dispatch({ type: GO_BACK }),
         goForward: () => dispatch({ type: GO_FORWARD }),
```
desbo, line 51: should we also track the progress to the next step here?

Gruntfile.js
```diff
@@ -61,7 +61,8 @@ module.exports = function (grunt) {
                 },
 
                 entry: {
-                    main: "./src/main",
+                    contributePage: 'src/contributePage',
+                    thankYouPage: 'src/thankYouPage',
```
desbo, line 65: good call on separating these

assets/javascripts/src/modules/analytics/ga.es6
desbo: I'd use the Redux store. At initialisation you can dispatch an action to either enable or disable tracking, and then use `store.getState()` to read that state whenever you need it.

assets/javascripts/src/reducers/gaTracking.es6
```diff
@@ -0,0 +1,12 @@
+import { TRACKING } from 'src/actions';
+
+const initialState = [false, false, false, false];
```
desbo: What's this state for?

alexduf: There's one boolean per step in our checkout process. This is used to debounce tracking events.

assets/javascripts/src/components/Main.jsx
alexduf: as we talked about I don't think it's necessary as making it compatible for mobile covers the "next" button only displayed on desktops

assets/javascripts/src/modules/analytics/ga.es6
AWare: Hmm, maybe initialise an no op function in ga.es6 so that if it's called without being initialised then it doesn't do anything. Rather than storing another proxy for the do not track  status. I can backport this into the membership ga js. (which is the same)

alexduf: I went with the redux option in the end

assets/javascripts/src/actions.es6
```diff
@@ -46,6 +46,7 @@ export function submitPayment(dispatch) {
             if (response.response.ok) dispatch({ type: PAYMENT_COMPLETE, response: response.json })
             else dispatch({ type: PAYMENT_ERROR, kind: 'card', error: response.json })
         })
+        .then(trackPayment(state.card.amount, state.data.currency.code))
```
desbo: the `PAYMENT_COMPLETE` dispatch above will update the browser address, that might happen before the GA event goes out. I'd move `trackPayment` to happen before the dispatch

desbo: in fact we should probably create an action for paymentComplete, which does both the GA tracking and redirect, and use that here and with paypal

alexduf: thanks, I fixed it

desbo (approved):
nice!

PR #129, opened 2016-09-19
Fix postcode submission

desbo:
The casing on the postcode field name was inconsistent, which was preventing this bit of data going through.
@alexduf


alexduf:
Looks good! :+1: (when build fixed)

PR #130, opened 2016-09-19
Add session stickyness to reduce problems during deployments

alexduf:
When we deploy we currently have a high risk of serving a borken page to our clients.

This is because during deplpyment two competing versions of the assets may be served by different servers.
For instance, a new version of the HTML could be served by the new version of the app, while older servers may serve 404 when trying to served the corresponding hashed assets.

This will limit the problem but won't solve it. It ensures a client sticks to the same AWS instance.

It's been tested on CODE


alexduf:
@rtyley @desbo

desbo (approved):

PR #131, opened 2016-09-19
Paypal postcode fix

pvighi:
Try getting the postcode from the billing and shipping address


alexduf (approved):
:+1: looks very sensible to me

PR #132, opened 2016-09-19
Make a note about having membership crendentials

philwills:
In case like me you hadn't got a fresh set yet.


alexduf:
Thanks! :+1:

alexduf (approved):

PR #135, opened 2016-09-19
Add resolver for identity-cookie

philwills:
In the medium term if we need this dependency we should publish it properly, but in the mean time this stops things breaking for those without it in their ivy cache.


AWare (approved):
Nice catch.

PR #137, opened 2016-09-20
Add SBT devrun command to readme

desbo:
A minor fix, but actually quite an important step in running locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,11 @@ For development you'll also need the following commands:
 npm run compile
 ```
 
+**Run the Play app**
+```
+sbt devrun
+```
+
 **Watch files for changes**
 
 ```
@@ -111,7 +116,6 @@ npm test
 ```
 
 #### Scala tests
-
 ```
 sbt test
 ```
--- a/README.md
+++ b/README.md
@@ -51,22 +51,6 @@ For development you'll also need the following commands:
 npm run compile
 ```
 
-**Run the Play app**
-```
-sbt devrun
-```
-
-**Watch files for changes**
-
-```
-npm run watch
-```
-
-This runs [webpack-dev-server](https://webpack.github.io/docs/webpack-dev-server.html) on port 7777, proxying requests to 9111. In this case, the bundled JS is stored in memory, which allows for much faster recompilation on changes, but the main.js file won't be created. You can also access localhost:7777/webpack-dev-server/\<route\> for a live reload environment. 
-
-If you want to compile main.js as a file, you can use `npm run compile` on its own and access the Play application directly.
-
-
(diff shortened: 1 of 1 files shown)


alexduf:
I think it's a good idea, but I would add that a bit lower in the "Run" section. Something saying if you want to run it manually blahblah ?

desbo:
Good call, I've reshuffled things a bit in the readme

README.md
```diff
@@ -99,6 +88,23 @@ This will start the Play application, which usually listens on port `9111`. Maki
 
 To make the site reachable as `contributions.thegulocal.com` (necessary for register/sign-in functionality) you then need to make sure NGINX is configured and running as described in [`/nginx/README.md`](./nginx/README.md).
 
+
+### Running the play app manuall
+```
+sbt devrun
```
AWare, line 94: ./start-frontend.sh?

README.md
```diff
@@ -99,6 +88,23 @@ This will start the Play application, which usually listens on port `9111`. Maki
...
+```
+
+### Watch files for changes
+
+```
+npm run watch
+```
+
+This runs [webpack-dev-server](https://webpack.github.io/docs/webpack-dev-server.html) on port 7777, proxying requests to 9111. In this case, the bundled JS is stored in memory, which allows for much faster recompilation on changes, but the main.js file won't be created. You can also access localhost:7777/webpack-dev-server/\<route\> for a live reload environment.
```
AWare, line 103: This approach doesn't have ssl, which will be more important when we do more on the thank you page as that breaks play sessions. There's a little work remaining to properly proxy this securely, but if you add giraffe.thegulocal.com to HOSTS it mostly works.

AWare (changes requested):

README.md
alexduf: yeah I think there's a broader conversation to have on how to start the dev server now we have webpack.
Could that be integrated with sbt? or should we have a local architecture with Nginx -> webpack -> play ?

README.md
desbo: look 10 lines up

README.md
desbo: Yeah, it seems a little confusing that there's nginx hosts for `contribute.thegulocal.com` and `giraffe.thegulocal.com`. We should rework how we run things locally for sure, and also tidy up the nginx config and decide on one hostname to use for development – that'll give us a much cleaner environment to document.

README.md
AWare: My bad. Sorry.

desbo:
I'm going to close this because I think there's a fair bit more work to do with the dev workflow, so we might as well re-do the docs at the same time as that.

PR #138, opened 2016-09-20
Use fetch instead of curl for Krux request

desbo:
We've had a [report](https://sentry.io/the-guardian/contributions-js/issues/147346411/) of an error in the Krux request. I'm not sure how curl handles exceptions, but `fetch` should be better, and there's no reason not to upgrade this anyway.


alexduf (approved):
Cool :+1:

PR #141, opened 2016-09-21
Remove MessageCopyTest

desbo:
I don't think this is being used at all, although I could be wrong.
Any ideas @AWare?


AWare:
It was a nasty hack before we implemented campaign codes. Is anything coming through with different variants?

desbo:
The variant gets selected and added as an attribute on the React top-level element, but we don't do anything with it. It's not being rendered at all as far as I can tell.

alexduf:
:+1:

AWare (approved):

PR #142, opened 2016-09-21
Add recurring payments A/B test

desbo:
Note that this isn't the whole test, just the back-end part. It also make a couple of changes to make VariantData optional, since in this case, I don't need any data, just knowing the variant is enough.

The rest of the work will be rendering the appropriate React components based on the variant which will follow in another PR.

### DIFF
--- a/app/views/support/ABTest.scala
+++ b/app/views/support/ABTest.scala
@@ -1,7 +1,5 @@
 package views.support
 
-
-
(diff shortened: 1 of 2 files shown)


rtyley:
cc @alexduf , he wants to know about it!

alexduf (approved):
:+1:

PR #143, opened 2016-09-22
Fix for paypal marketing opt in

pvighi:
PayPal returns different emails for the same payer in different api calls so we should make sure to use the exact same email we used before when adding the marketing opt in value

### DIFF
--- a/app/controllers/PaypalController.scala
+++ b/app/controllers/PaypalController.scala
@@ -42,27 +42,13 @@ class PaypalController(
     val idUser = IdentityUser.fromRequest(request).map(_.id)
 
     paypalService.executePayment(paymentId, payerId) match {
-      case Right(executedPayment) =>
-        paypalService.storeMetaData(paymentId, chosenVariants, cmp, intCmp, ophanId, idUser)
-        getEmail(executedPayment) match {
-          case Some(email) => redirectWithCampaignCodes(postPayUrl).withSession(request.session + ("email" -> email))
-          case None =>
-            Logger.error(s"No email returned from Paypal payment execution paymentId=$paymentId")
-            redirectWithCampaignCodes(thanksUrl)
+      case Right(_) =>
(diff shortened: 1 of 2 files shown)


app/controllers/PaypalController.scala
```diff
@@ -42,29 +42,15 @@ class PaypalController(
...
-        paypalService.storeMetaData(paymentId, chosenVariants, cmp, intCmp, ophanId, idUser)
-        getEmail(executedPayment) match {
-          case Some(email) => redirectWithCampaignCodes(postPayUrl).withSession(request.session + ("email" -> email))
-          case None =>
-            Logger.error(s"No email returned from Paypal payment execution paymentId=$paymentId")
-            redirectWithCampaignCodes(thanksUrl)
+      case Right(_) =>
+        paypalService.storeMetaData(paymentId, chosenVariants, cmp, intCmp, ophanId, idUser) match {
+          case Right(savedData) => redirectWithCampaignCodes(postPayUrl).withSession(request.session + ("email" -> savedData.contributor.email))
+          case Left(_) => redirectWithCampaignCodes(thanksUrl)
```
alexduf, line 48: We should probably trace an error here

app/services/PaypalService.scala
```diff
@@ -115,14 +115,16 @@ class PaypalService(config: PaypalApiConfig, contributionData: ContributionData)
...
+
   def storeMetaData(
     paymentId: String,
     chosenVariants: ChosenVariants,
     cmp: Option[String],
     intCmp: Option[String],
     ophanId: Option[String],
     idUser: Option[String]
-  ): Either[String,String ] = {
+  ): Either[String,SavedContributionData ] = {
```
alexduf: spacing

alexduf (changes requested):

alexduf (approved):

PR #144, opened 2016-09-22
tracking reducer reimplemented so that it does not mutate state

Mullefa:
@desbo the current implementation of `trackingReducer()` mutates state which is not cool according to the redux docs. I've reimplemented it (with my limited knowledge of JS!).

### DIFF
--- a/assets/javascripts/src/reducers/gaTracking.es6
+++ b/assets/javascripts/src/reducers/gaTracking.es6
@@ -5,22 +5,13 @@ const initialState = {
     steps: [false, false, false, false]
 };
 
-// return a copy of `array` with element `i` having its value replaced with `value`
-const replace = (array, i, value) => {
-    let newArray = array.slice();
-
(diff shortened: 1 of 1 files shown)


assets/javascripts/src/reducers/gaTracking.es6
```diff
@@ -5,14 +5,30 @@ const initialState = {
...
-        return state;
-    } else if (action.type == GA_ENABLED) {
-        state.enabled = action.enabled;
-        return state;
-    } else {
-        return state;
+    switch (action.type) {
+        case TRACK_STEP:
+            return Object.assign({}, state, {
+                steps: replace(state.steps, action.step, true)
```
desbo: you could try using `state.steps.map` here, in JS `map` gets the value and index as params so may be a little cleaner than a separate function

desbo (approved):
awesome

PR #147, opened 2016-09-23
move the jdbc calls to a different thread pool

alexduf:
JDBC calls are currently blocking. It doesn't cause any issue given our current load, but I hope we'll soon get a load big enough that would force us to fix it.

I also noticed our stripe client and probably the paypal client are blocking as well. That will come later.

cc @rtyley


rtyley:
This might be a bit crazy, but I was wondering if you'd considered ditching JDBC and using something like https://github.com/mauricio/postgresql-async instead? I've never used it, just thought it might be nice to use a natively async client.

pvighi (approved):
looks alright to me

PR #148, opened 2016-09-23
Only allocate each user to one test

desbo:
This is an attempt to avoid test conflicts by never assigning more than one test to any user. It works like this...

for new users (no cookies):
1. assign user a random test ID, stored in a cookie
2. use that ID to select one of the currently active tests, and randomly pick a variant
3. store the result of that selection in a cookie

for returning users (test/variant cookies), their cookie is used to determine the test and variant, so people aren't reallocated on repeat visits.

When tests are deactivated (by removing them from the list of tests) the cookie associated with that test gets dropped 👋 

This could probably be tidied up/made much better so feel free to let loose in the comments.

### DIFF
--- a/app/controllers/Giraffe.scala
+++ b/app/controllers/Giraffe.scala
@@ -129,13 +129,11 @@ class Giraffe(paymentServices: PaymentServices) extends Controller with Redirect
     val maxAmountInLocalCurrency = MaxAmount.forCurrency(countryGroup.currency)
     val creditCardExpiryYears = CreditCardExpiryYears(LocalDate.now.getYear, 10)
 
-    val inactiveTestsInCookies = request.cookies.filter(_.name.contains(Test.CookiePrefix))
-      .filterNot(c => c.value == variant.testSlug)
-      .flatMap(c => Test.allTests.find(_.slug.equalsIgnoreCase(c.name)))
+    val testsInCookies = request.cookies.filter(_.name.contains(Test.CookiePrefix)) map(_.name)
 
     Ok(views.html.giraffe.contribute(pageInfo, maxAmountInLocalCurrency, countryGroup, variant, cmp, intCmp, creditCardExpiryYears, errorMessage))
+      .discardingCookies(testsInCookies.toSeq map(DiscardingCookie(_)): _*)
       .withCookies(Test.testIdCookie(mvtId), Test.variantCookie(variant))
-      .discardingCookies(inactiveTestsInCookies.toSeq map (t => DiscardingCookie(Test.cookieName(t))): _*)
   }
 
   def thanks(countryGroup: CountryGroup) = NoCacheAction { implicit request =>
--- a/app/controllers/Giraffe.scala
+++ b/app/controllers/Giraffe.scala
@@ -110,11 +110,13 @@ class Giraffe(paymentServices: PaymentServices) extends Controller with Redirect
 
   def contribute(countryGroup: CountryGroup, error: Option[PaymentError] = None) = NoCacheAction { implicit request =>
     val mvtId = Test.testIdFor(request)
+    val variant = Test.getContributePageVariant(countryGroup, mvtId, request)
+
     val errorMessage = error.map(_.message)
     val stripe = paymentServices.stripeServiceFor(request)
     val cmp = request.getQueryString("CMP")
     val intCmp = request.getQueryString("INTCMP")
-    val variant = Test.getContributePageVariant(countryGroup, mvtId, request)
+
     val pageInfo = PageInfo(
       title = "Support the Guardian | Contribute today",
       url = request.path,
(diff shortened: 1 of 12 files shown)


desbo:
Also, the work to support the payment methods in React is pretty dirty but once @pvighi makes PayPal  a standard feature I'll update it.

app/controllers/Giraffe.scala
```diff
@@ -124,8 +126,14 @@ class Giraffe(paymentServices: PaymentServices) extends Controller with Redirect
...
-    Ok(views.html.giraffe.contribute(pageInfo, maxAmountInLocalCurrency, countryGroup, chosenVariants, cmp, intCmp, creditCardExpiryYears, errorMessage))
-      .withCookies(chosenVariants.variants.map(Test.createCookie):_*)
+    val response = Ok(views.html.giraffe.contribute(pageInfo, maxAmountInLocalCurrency, countryGroup, variant, cmp, intCmp, creditCardExpiryYears, errorMessage))
+
+    val responseWithTests = (for {
+      testId <- request.cookies.get(Test.TestIdCookieName)
+      variantSlug <- request.cookies.get(Test.cookieName(variant))
+    } yield response) getOrElse response.withCookies(Test.testIdCookie(mvtId), Test.variantCookie(variant))
+
+    responseWithTests.discardingCookies(Test.allTests.filterNot(t => t.slug == variant.testSlug) map(t => DiscardingCookie(Test.cookieName(t))): _*)
```
alexduf: I feel like you could simplify the logic by discarding all the test cookies first, then adding the right cookies?

app/services/PaypalService.scala
```diff
@@ -137,7 +138,7 @@ class PaypalService(config: PaypalApiConfig, contributionData: ContributionData)
         created = created,
         email = payerInfo.getEmail,
         ophanId = ophanId,
-        abTests = chosenVariants.asJson,
+        abTests = Json.toJson(variant),
```
alexduf: I know it's going to sound crazy, but I think if we want the DB to be backward compatible, we need that value to be wrapped into an array. I think something like that should be possible

```
Json.Array(Json.toJson(variant))
```

or

```
Json.toJson(Seq(variant))
```

app/views/support/ABTest.scala
```diff
@@ -142,41 +140,45 @@ object RecurringPaymentTest extends TestTrait {
...
   }
 
-  def createCookie(variant: Variant): Cookie = {
-    Cookie(variant.testSlug+"_GIRAFFE_TEST", variant.variantSlug, maxAge = Some(604800))
+  def pickTest[A](request: Request[A], userId: Int): TestTrait = {
+    def pickByQueryString = allTests.find(t => request.getQueryString(t.slug).nonEmpty)
+    def pickByCookie = allTests.find(t => request.cookies.get(s"$CookiePrefix.${t.slug}").nonEmpty)
+    def pickRandomly = allTests(userId % allTests.length)
+
+    pickByQueryString orElse pickByCookie getOrElse pickRandomly
```
alexduf, line 150: 👍 I love when it gets that readable!

alexduf (changes requested):
I think the compatibility at the db level needs to be addressed before merging

app/controllers/Giraffe.scala
alexduf: this comment is completely optional BTW

app/services/PaypalService.scala
desbo: No, that makes sense – I actually wrap it in a `Seq` in the template, since all the JS stuff is written to handle arrays, but that's something that should happen further up.

app/views/support/ABTest.scala
Mullefa: Small details, but I don't think `pickByQueryStringOrCookie` needs a type parameter. Also, the code can be made even more readable (yay!) by using the `contains` method:

``` scala
test.variantsByCountry(countryGroup).find(variant => search.contains(variant.variantSlug))
```

app/controllers/Giraffe.scala
Mullefa: I agree, or kept as is, I think it'd be clearer if an explicit check was made for whether the options were defined (instead of using a for loop).

desbo: I've had a play about with this, but I don't know if we should discard all of the cookies first. The result of that is sending the client a `set-cookie` header on every response, even if that cookie is for the same test as the one they already had set. In that case, there's no need to do the `set-cookie`, you can just let the client keep the cookie as-is.  

It seems inefficient to constantly send `set-cookie` headers instead of only sending them when they're needed, which is how it works currently. What do you think?

assets/javascripts/src/components/Main.jsx
```diff
@@ -130,8 +129,8 @@ class Main extends React.Component {
             </MediaQuery>
         </div>
     }
+
     renderSummary(visible) {
```
pvighi, line 133: There is some useless code left over by my previous pull request.
There is no need for renderSummary anymore since both the mobile and desktop cases are the same now.
I can fix this in another pull request or I can push a fix here if you want

desbo: Ah cool, pushing it here works for me

app/controllers/Giraffe.scala
alexduf: I think both are fine. True it would set the cookie, but on a normal workflow you only load that page once anyway :D
No need to change anything though

assets/javascripts/src/components/Main.jsx
alexduf: I'd put that on a separate PR to keep the change as atomic as possible

alexduf (approved):

app/controllers/Giraffe.scala
desbo: True. OK, after another look at it I think what I have is reasonably clean, it does set the cookies each time, but that's probably not such a big deal.

PR #150, opened 2016-09-25
Minor fixes

kant:
Links leads to nowhere


desbo (changes requested):
Thanks @kant! Looks like these links were carried over from an old repo. I don't think there's any need for that part of the readme, since `docs` doesn't exist. Would you mind removing the whole `additional documentation` section?

PR #151, opened 2016-09-26
Remove the going to second step part of the amount test

jranks123:
Previously, when the user came in from the contributions ask with the buttons, they were directed top the second step of the process. This is no longer a requirement.


alexduf:
👍

PR #153, opened 2016-09-26
Use a different thread pool for PayPal calls

pvighi:


app/controllers/PaypalController.scala
```diff
@@ -46,13 +45,13 @@ class PaypalController(
     val paypalService = paymentServices.paypalServiceFor(request)
     val idUser = IdentityUser.fromRequest(request).map(_.id)
 
-    paypalService.executePayment(paymentId, payerId) match {
-      case Right(_) =>
+    paypalService.executePayment(paymentId, payerId).value flatMap {
+      case Xor.Right(_) =>
```
alexduf: I think this nested pattern match needs to be re-written as it gets hard to follow. I know it's not the point of this PR but that would be nice

app/services/PaypalService.scala
```diff
@@ -88,34 +97,28 @@ class PaypalService(config: PaypalApiConfig, contributionData: ContributionData)
...
-      val createdPayment = payment.create(apiContext)
-      createdPayment.getLinks.asScala
-    } match {
-      case Success(links) =>
-        val approvalLink = links.find(_.getRel.equalsIgnoreCase("approval_url")).map(l => Right(addUserActionParam(l.getHref)))
-        approvalLink.getOrElse(Left("No approval link returned from paypal"))
-      case Failure(exception) => Left(exception.getMessage)
+    val res = asyncExecute {
+      val links = payment.create(apiContext).getLinks.asScala
+      links.find(_.getRel.equalsIgnoreCase("approval_url")).map(l => addUserActionParam(l.getHref))
```
alexduf: I think you need to separate your concerns here. You're both searching for something and mapping on the value.

I would wrap only `payment.create(apiContext)` in the async block and start from there

app/services/PaypalService.scala
```diff
@@ -88,34 +97,28 @@ class PaypalService(config: PaypalApiConfig, contributionData: ContributionData)
...
-      case Success(payment) =>
-        if (payment.getState.toUpperCase != "APPROVED") {
-          Left(s"payment returned with state: ${payment.getState}")
-        } else {
-          Right(payment)
-        }
-      case Failure(exception) =>
-        Logger.error("Unable to execute payment", exception)
-        Left(exception.getMessage)
+    val paymentResponse = asyncExecute(payment.execute(apiContext, paymentExecution)).value map {
```
alexduf: If you want to map over the `Xor`, according to the doc you should be able to use [transform](http://typelevel.org/cats/api/index.html#cats.data.XorT@transform[C,D]%28f:cats.data.Xor[A,B]=%3Ecats.data.Xor[C,D]%29%28implicitF:cats.Functor[F]%29:cats.data.XorT[F,C,D]) here like that:

```
asyncExecute(payment.execute(apiContext, paymentExecution)).transform {
  case Xor.Right(blah) => some Xor result
}
```

and then there's no need for your final `XorT(paymentResponse)`

app/services/PaypalService.scala
```diff
@@ -88,34 +97,28 @@ class PaypalService(config: PaypalApiConfig, contributionData: ContributionData)
...
-      case Failure(exception) => Left(exception.getMessage)
+    val res = asyncExecute {
+      val links = payment.create(apiContext).getLinks.asScala
+      links.find(_.getRel.equalsIgnoreCase("approval_url")).map(l => addUserActionParam(l.getHref))
+    }.value map {
+      case Xor.Right(None) => Xor.left("No approval link returned from PayPal")
+      case Xor.Right(Some(link)) => Xor.right(link)
+      case Xor.Left(error) => Xor.left(error)
     }
+    XorT(res)
```
alexduf: see comment below

alexduf (changes requested):

app/controllers/PaypalController.scala
alexduf: this one is optional, just nice to have

app/services/PaypalService.scala
pvighi: that is much better, I'll refactor it like that

app/services/PaypalService.scala
```diff
@@ -66,59 +66,68 @@ class PaypalService(config: PaypalApiConfig, contributionData: ContributionData)
...
-    val res = asyncExecute {
-      val links = payment.create(apiContext).getLinks.asScala
-      links.find(_.getRel.equalsIgnoreCase("approval_url")).map(l => addUserActionParam(l.getHref))
-    }.value map {
-      case Xor.Right(None) => Xor.left("No approval link returned from PayPal")
-      case Xor.Right(Some(link)) => Xor.right(link)
+    asyncExecute{
+      paymentToCreate.create(apiContext)
+    } transform {
+      case Xor.Right(createdPayment) => Xor.fromOption(getAuthLink(createdPayment), "No approval link returned from PayPal")
```
alexduf, line 105: looking much better 👍

app/services/PaypalService.scala
```diff
@@ -66,59 +66,68 @@ class PaypalService(config: PaypalApiConfig, contributionData: ContributionData)
...
       case Xor.Left(error) => Xor.left(error)
     }
-    XorT(res)
+
   }
 
-  private def addUserActionParam(url:String) = Uri.parse(url).addParam("useraction", "commit").toString
+  private def getAuthLink(payment: Payment) = {
```
alexduf, line 110: [Optional] That would be nice to add Option[Uri] as the return type here

pvighi: I thought about that but I wasn't sure if we wanted to expose that Uri class on the public interface of the service.

alexduf (approved):
👍

PR #155, opened 2016-09-26
Reduced checkout (WIP)

AWare:
aka 2 step giraffe
This removes the middle page as a test.

Could do with better tracking tags. Will update with suggestions.

### DIFF
--- a/app/views/support/ABTest.scala
+++ b/app/views/support/ABTest.scala
@@ -38,7 +38,7 @@ object Variant {
       (__ \ "variantSlug").write[String] and
       (__ \ "weight").write[Double] and
       (__ \ "data").writeNullable[VariantData]
-    )(v => (v.testName, v.testSlug, v.variantName, v.variantSlug, v.weight, v.data))
+    ) (v => (v.testName, v.testSlug, v.variantName, v.variantSlug, v.weight, v.data))
 }
 
 trait TestTrait {
@@ -54,7 +54,7 @@ trait TestTrait {
     def filterVariants(countryGroup: CountryGroup) = variants.list.filter(_.matches(countryGroup)).toSet
 
     CountryGroup.allGroups.map(country => country -> filterVariants(country)).toMap
-  }
+    }
 
   val variantRangesByCountry: Map[CountryGroup, Seq[(Double, Variant)]] = {
    def addRanges(filteredVariants: Set[Variant]): Seq[(Double, Variant)] = {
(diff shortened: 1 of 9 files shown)


AWare:
fixed a merge conflict, will double check in morning

AWare:
:-1: need to collect marketing prefs

app/views/support/ABTest.scala
```diff
@@ -66,12 +73,26 @@ trait TestTrait {
...
 }
 
+object ReducedCheckoutTest extends TestTrait {
+  def name = "ReducedCheckoutTest"
+
+  def slug = "2sc"
+
+  def variants = NonEmptyList(
+    makeVariant("control", "c", 10, Some(BooleanVariantData(false))),
+    makeVariant("test", "t", 1, Some(BooleanVariantData(true)))
```
desbo: Because you're only providing `true` and `false` as the variant data, you can just use `None` here, and avoid having to specify the json formatter. On the client you can just check which variant you're in rather than reading the data.

AWare: Neat.

assets/javascripts/src/components/form-wrapper/MobileWrapper.jsx
```diff
@@ -9,6 +9,7 @@ import {ALL_PAGES, PAGES} from 'src/constants';
 export default class MobileWrapper extends React.Component {
 
     render() {
+        console.log(this.props.reducedCheckout);
```
desbo: 🔭

AWare: 🔥🚀🔥

assets/javascripts/src/components/pages/Payment.jsx
```diff
@@ -80,7 +80,17 @@ export default class Payment extends React.Component {
                             className="center-text"
                             outerClassName="half-width"
                             required/>
+
             </div>
+            {this.props.reducedCheckout && <InputField label="Email" id="email2" type="email" value={this.props.details.email}
+                                                      onChange={ event => this.props.updateDetails({ email: event.target.value })}
+                                                      required />}
+            {this.props.reducedCheckout && <div className="giraffe-checkbox">
```
desbo, line 88: Might be cleaner to wrap all of this in one condition

AWare: This would need both elements to share a root node though, which I'm not keen on.

desbo (changes requested):
I think we should get rid of the boolean data stuff before merging

desbo (approved):

AWare:
Will rebase and merge when I get the numbers from Jack.

PR #156, opened 2016-09-27
Basic Auth

alexduf:
Basic Auth is going to be needed for stripe's webhooks.
This PR is an effort to make the webhook one look a bit smaller, as such it doesn't provide any change to the main logic, but just make a new feature available.

It also fixes the riff-raff deprecation warning we have during deployment


alexduf:
actually I'll use a query parameter here 😞 
Over-engineering FTW!

PR #157, opened 2016-09-27
be nicer when ga is blocked

alexduf:
If the ga script is blocked I imagine the window.ga function won't be defined. This fix should fix a few of the errors we see in sentry.


AWare (approved):
:+1:

PR #158, opened 2016-09-27
Stripe webhook

alexduf:
This PR ads support for the stripe webhook in the contributions-frontend.

Once merged we need to reconfigure the webhook on the stripe side.

I've updated the conf of all environments to support the webhook secret

### DIFF
--- a/app/controllers/Giraffe.scala
+++ b/app/controllers/Giraffe.scala
@@ -2,24 +2,30 @@ package controllers
 
 import java.lang.Math.min
 import java.time.LocalDate
+import java.util.UUID
 
 import actions.CommonActions.NoCacheAction
+import cats.data.XorT
 import com.gu.i18n.CountryGroup._
 import com.gu.i18n._
 import com.gu.stripe.Stripe
+import com.gu.stripe.Stripe.Charge
 import com.gu.stripe.Stripe.Serializer._
 import com.netaporter.uri.dsl._
 import configuration.Config
+import models.SavedContributionData
+import org.joda.time.DateTime
 import play.api.data.Forms._
 import play.api.data.format.Formatter
-import play.api.data.{FieldMapping, Form, FormError}
+import play.api.data.{Form, FormError}
 import play.api.libs.concurrent.Execution.Implicits._
-import play.api.libs.json.{JsArray, JsString, Json}
+import play.api.libs.json.Json
 import play.api.mvc._
 import services.PaymentServices
 import utils.MaxAmount
 import utils.RequestCountry._
 import views.support._
+
 import scala.concurrent.Future
 
 class Giraffe(paymentServices: PaymentServices) extends Controller with Redirect {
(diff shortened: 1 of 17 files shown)


pvighi:
Can we save some of the data before the webhook call instead of sending everything to stripe and back?

app/models/PaymentHook.scala
```diff
@@ -56,10 +67,45 @@ case class PaymentHook(
...
+    cardCountry = None,
+    amount = paypalHook.amount,
+    convertedAmount = None,
+    status = paypalHook.status,
+    email = None
+  )
 
-  implicit val paypalReader = new Reads[PaymentHook] {
+  def fromStripe(stripeHook: StripeHook, convertedAmount: BigDecimal): PaymentHook = PaymentHook(
+    contributionId = UUID.nameUUIDFromBytes(stripeHook.paymentId.getBytes),
```
pvighi: the PayPal version of this generates the contribution Id using ContributionIdGenerator before executing the payment. We should probably use that class to generate this to make PayPal and stripe contribution ids are generated in the exact same way. 
We could even generate the id in a common part of the flow for both types of payments

pvighi (changes requested):

alexduf:
I modified this PR to adopt the same workflow as paypal (store data as soon as we have it). It wasn't too bad in the end.
I left all the metadata on the stripe charge, so that if we were to have any problem with the database we could rebuild it from the stripe transactions

app/controllers/Giraffe.scala
```diff
@@ -148,40 +154,71 @@ class Giraffe(paymentServices: PaymentServices) extends Controller with Redirect
...
-      res.map { charge =>
-        Ok(Json.obj("redirect" -> redirect))
-          .withSession(chargeId -> charge.id)
-      }.recover {
-        case e: Stripe.Error => BadRequest(Json.toJson(e))
-      }
-    })
+    val variant = Test.getContributePageVariant(countryGroup, Test.testIdFor(request), request)
+
+    val contributionId = UUID.randomUUID()
```
pvighi: could you use the ContributionIdGenerator object ? the only reason for it to exist is to make sure PayPal and Stripe generate contribution ids in the same way.  It's not a big deal but if it's only going to be used from PayPal I think it's better to just delete the object

alexduf: The reason it exists is to link the meta-data of a transaction with a transaction as they're stored in a different table.
I could make a contributionId type but I'm not sure we'd gain much from it

pvighi: What I mean is that there is a very simple [ContributionIdGenerator object](https://github.com/guardian/contributions-frontend/blob/master/app/utils/ContributionIdGenerator.scala) defined with a method getNewId that does exactly this. 
The only reason I used that in paypal instead of just calling UUID.randomUUID was so that both paypal and stripe could call the same code to generate the contribution id since it's the same field.
If you don't think that is useful we can leave this like it is and later delete ContributionIdGenerator and make the paypal code call UUID.randomUUID directly

alexduf: Ah I see! I'll fix that then

pvighi (approved):

pvighi:
Other than that tiny detail with the contribution id  it looks fine by me, you might want someone who knows more about the original stripe webhook to take a look

jranks123:
👍

PR #159, opened 2016-09-27
Add recurring payments test

desbo:
A test for recurring payments!

sep 27 2016 17 20

In addition to the changes required for those visuals, this fires an Ophan `complete` event for the test so we can measure the results in Abacus, and also fires a GA event with a value so we can see the average amount people are choosing to pay each month. This doesn't include currency, so the results will be far from perfect, but should be indicative.

This ended up being much bigger than I expected, apologies for the massive PR :(

### DIFF
--- a/assets/javascripts/src/actions.es6
+++ b/assets/javascripts/src/actions.es6
@@ -28,6 +28,7 @@ export const GA_ENABLED = "GA_ENABLED";
 
 export const SET_RECURRING = "SET_RECURRING";
 export const SET_RECURRING_NOTIFIED = "SET_RECURRING_NOTIFIED";
+export const DISMISS_RECURRING_NOTIFICATION = "DISMISS_RECURRING_NOTIFICATION";
 
 export function submitPayment(dispatch) {
     const state = store.getState();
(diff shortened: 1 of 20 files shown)


AWare:
If you tab through the form and select monthly -> ok the focus drops off the end of the form.

AWare:
The 25 is hoverable and the default selection changes on one-off. Also, might be worth letting the user just click on an amount and have it autoselect one-off? 

image
image

desbo:
Thanks @AWare, I'll get to those now. Ben and Ali were pretty keen on the form being totally inactive until one of the options has been selected, and I agree, so I don't think we should make the amounts clickable until we've chosen one. Nick H is working on a bit of new copy for "How often?" that should make it clearer you have to click one of the options.

AWare:
@desbo yeah, that makes sense.

tomverran:
You could always link people who want to make recurring contributions to membership, at least for now!

desbo:
cheers @tomverran, I've added that.

desbo:
@AWare  squashed this mother down a bit! can I get a ✅ ?

AWare (approved):
🆗

PR #161, opened 2016-09-29
Remove some membership cruft from the README

philwills:


README.md
```diff
@@ -133,8 +131,10 @@ S3 bucket, and if private credentials need adding or updating, they need to be u
 
 You can download and update credentials like this
 
-    aws s3 cp s3://contributions-private/DEV/contributions.private.conf /etc/gu
-    aws s3 cp /etc/gu/contributions.private.conf s3://contributions-private/DEV/
+```
+sudo aws s3 cp s3://contributions-private/DEV/contributions.private.conf /etc/gu/ --profile membership
```
alexduf, line 135: I think it's probably better practice to leave the `sudo` out, and let the developer handle the file. Typically there's no need to run `aws` as root if you can manually move the file afterwards (But that's me being paranoid and pedantic, I'll go put my tinfoil hat now if you don't mind)

alexduf (approved):
Please feel free to disregard previous comment and merge 👍

README.md
AWare: We could be good and use ~/.gu instead?

alexduf: yeah personally I put everything in ~/.gu, and create symlinks when there's no other choice (nginx for instance)

philwills: Yea, I don't like dumping anything accessible only by `sudo`.  I'll probably do a separate PR around that. Once I have something working and can confirm that putting things in `~/.gu` works fine too.

PR #162, opened 2016-09-29
fix checkbox label clickability

alexduf:
we can't click on the label to uncheck the checkbox right now, which require the user to be super precise with its mouse.

This PR fixes it.


philwills (approved):
👍 👍

PR #163, opened 2016-09-29
We should use INTCMP to track redirections between Guardian propertie…

paulbrown1982:
We should use INTCMP to track redirections between Guardian properties, not CMP - which is for externally-driven marketing campaigns.

cc @ajosephides please correct me if I'm wrong?
cc @AWare @desbo


desbo (approved):
My bad.

ajosephides:
@paulbrown1982  - correct.
INTCMP for internal campaigns.
CMP will be phased out with Omniture. These external campaigns will be replaced by the google analytics recognised set (5 or which 3 mandatory) UTM parameters.

PR #164, opened 2016-09-30
Second attempt at 6 value test

philwills:
Updated with new design from Ben

### DIFF
--- a/assets/javascripts/src/components/InputField.jsx
+++ b/assets/javascripts/src/components/InputField.jsx
@@ -4,7 +4,7 @@ export class AmountInput extends React.Component {
     render() {
         const { inputAmount, symbol, refFn, ...props } = this.props;
 
-        return <span className={(this.props.small ? 'contribute-controls__input-small':'contribute-controls__input') + ' contribute-controls__input--amount input-text'}>
+        return <span className={'contribute-controls__input contribute-controls__input--amount input-text ' + (this.props.small ? 'contribute-controls__input-small':'')}>
             <span className={'symbol ' + (!!inputAmount ? 'active' : '')}>{symbol}</span>
             <input type="number"
                    ref={refFn.bind(this)} // create a reference to this element for validation (see: https://facebook.github.io/react/docs/more-about-refs.html)
--- a/assets/javascripts/src/components/InputField.jsx
+++ b/assets/javascripts/src/components/InputField.jsx
@@ -4,13 +4,14 @@ export class AmountInput extends React.Component {
     render() {
         const { inputAmount, symbol, refFn, ...props } = this.props;
 
-        return <span className="contribute-controls__input contribute-controls__input--amount input-text">
+        return <span className={(this.props.small ? 'contribute-controls__input-small':'contribute-controls__input') + ' contribute-controls__input--amount input-text'}>
             <span className={'symbol ' + (!!inputAmount ? 'active' : '')}>{symbol}</span>
             <input type="number"
                    ref={refFn.bind(this)} // create a reference to this element for validation (see: https://facebook.github.io/react/docs/more-about-refs.html)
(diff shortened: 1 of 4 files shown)


philwills:
image

**Warning**, this PR include CSS written by me, so please educate me on all the things I'm doing wrong.

assets/javascripts/src/components/InputField.jsx
```diff
@@ -4,13 +4,14 @@ export class AmountInput extends React.Component {
     render() {
         const { inputAmount, symbol, refFn, ...props } = this.props;
 
-        return <span className="contribute-controls__input contribute-controls__input--amount input-text">
+        return <span className={(this.props.small ? 'contribute-controls__input-small':'contribute-controls__input') + ' contribute-controls__input--amount input-text'}>
```
desbo: I would conditionally add an extra class here, `contribute-controls__input--amount--small`, leaving the classes as before if `props.small` is false.

assets/stylesheets/giraffe-react.scss
```diff
@@ -275,8 +275,14 @@ $control-spacing: 5px;
         }
     }
 
+    input.contribute-controls__small-input {
```
desbo: just the class name is enough

assets/javascripts/src/components/pages/Contribution.jsx
```diff
@@ -90,7 +90,7 @@ class FormWithRecurring extends React.Component {
                 : 'One-off amount';
 
 
-        return <div className={'contribute-controls contribute-fields contribute-controls--recurring ' + (this.props.amounts.length % 3 ? 'option-button__fours ' : 'option-button__three ')}>
+        return <div className="contribute-controls contribute-fields contribute-controls--recurring option-button__fours">
```
desbo: You probably want to make your changes to the `FormWithoutRecurring` component. This one will only be rendered if the user is in the recurring test, which means they won't be in the amounts test!

desbo: Oh, I see you've made them there as well, you don't need them here.

assets/javascripts/src/components/pages/Contribution.jsx
```diff
@@ -123,7 +125,7 @@ class FormWithRecurring extends React.Component {
 
 class FormWithoutRecurring extends React.Component {
     render() {
-        return <div className={'contribute-controls contribute-fields ' + (this.props.amounts.length % 3 ? 'option-button__fours' : 'option-button__three')}>
+        return <div className="contribute-controls contribute-fields option-button__fours">
```
desbo: It doesn't look like `option-button__fours` is defined anywhere in our css so it should be safe to get rid of it here

assets/javascripts/src/components/pages/Contribution.jsx
philwills: I changed both, just for consistency's sake, I don't believe it will have any effect.

assets/javascripts/src/components/pages/Contribution.jsx
philwills: Thanks, I've also got rid of the SASS for option-button__threes, as it's now unused.

assets/javascripts/src/components/InputField.jsx
philwills: Aha, I thought the width would compete with the flex-basis, but apparently not, thanks.

assets/javascripts/src/components/pages/Contribution.jsx
desbo: Fair enough - it's probably more to do with how I split out these components for A/B tests. I'm still trying to figure out the best way of doing it.

desbo (approved):

PR #167, opened 2016-09-30
[WIP] Initial code that will eventually let us send emails.

AWare:
🚨🚨 **THIS IS WORK IN PROGRESS AND IS BEING HANDED OVER DO NOT MERGE**🚨🚨

To send an email we need to:
- build the JSON enevlope to send to exact target
- add this to an SQS queue
- tell membership workflow about this queue

This commit makes way to 1 and 2.

The JSON that Exact Target requires needs to include:
- email
- creation date (a date time that will need  some formatting)
- amount
- card country (I presently am sending currency instead, we'll need to discuss this and get it changed to currency)
- name
- data extension (this is the name of the custom fieldset in exacttarget that we're using, if you located it, you can change membership common to add it like the others in their)

This is all straightforward with Stripe, but paypal needs an additional call to get the name and email.

Membership Workflow sends the final email; and this will need to know about the SQS queue and where in exacttarget to send it.

Other fun: To add the data extension to exacttarget, we'll need to update membership common. It's quite a jump, but I checked with Tom and he doesn't think he's changed anything we use.

### DIFF
--- a/app/configuration/Config.scala
+++ b/app/configuration/Config.scala
@@ -13,6 +13,7 @@ object Config {
   val contributeUrl = config.getString("contribute.url")
   val facebookAppId = config.getString("facebook.app.id")
   val idWebAppUrl = config.getString("identity.webapp.url")
+  val emailSQSQueue = config.getString("emailSQSQueue")
 
   val stage = config.getString("stage")
   val stageProd: Boolean = stage == "PROD"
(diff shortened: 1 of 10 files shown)


AWare:
@alexduf: here goes, hope it all makes sense

@tomverran: we're safe to update membership-common in giraffe?

PR #168, opened 2016-09-30
bump membership-common and make balance transaction optional

alexduf:
I bump the membership-common library to reflect a bugfix about stripe's webhook

`convertedAmount` is now optional as it is dependant on the presence of a `balance_transaction` coming from stripe.
`Balance_transaction` is `null` when we receive a failed payment from stripe webhook


alexduf:
cc @philwills @desbo

paulbrown1982:
cc @AWare in case this is relevant to your ExactTarget stuff.

desbo (approved):

PR #171, opened 2016-10-04
Marketing prefs

alexduf:
Ping the identity API to update the marketing prefs.

I'm using a fire-and-forget approach as in case of an error I don't want to stop the user's workflow. An error should trigger a sentry notification.

I also included a small refactor in the way we handle the creation of the payment service, and I added a type around the Identity user ID. It was a bit ambiguous as a String so I opted for a specific Type. Name suggestion welcome.


app/models/PaymentMode.scala
```diff
@@ -0,0 +1,18 @@
+package models
+
+sealed trait PaymentMode {
```
philwills, line 3: I keep meaning to try https://github.com/lloydmeta/enumeratum for this kind of thing.

app/services/IdentityService.scala
```diff
@@ -0,0 +1,30 @@
...
+  def request(path: String): WSRequest = {
+    wsClient.url(s"$identityUrl/$path")
+      .withHeaders("Authorization" -> s"Bearer $token")
+  }
+
+  def updateMarketingPreferences(userId: IdentityId, marketing: Boolean): Future[Boolean] = {
+    val payload = Json.obj("statusFields" -> Json.obj("receiveGnmMarketing" -> marketing))
+    request(s"user/${userId.id}").post(payload).map { response =>
+      response.status >= 200 && response.status < 300
+    } recover {
```
philwills, line 24: I think this makes sense for now, but I'd like to see us move to some form of asynchronous communication (Kinesies/SQS/SWF?) for this kind of thing, as we do really want it to happen, but it matters less exactly when it happens.

app/services/PaymentServices.scala
```diff
@@ -1,78 +1,67 @@
...
+        contributionData = contributionData,
+        identityService = identityService
+      )
+    }
+    PaymentMode.all.map(mode => mode -> stripeServiceFor(mode)).toMap
   }
 
-  def paypalServicesFor(paypalConfig: Config, contributionDataPerMode: Map[Mode, ContributionData])(ec: ExecutionContext): Map[Mode, PaypalService] = {
+  val paypalServices: Map[PaymentMode, PaypalService] = {
+    val paypalExecutionContext = actorSystem.dispatchers.lookup("contexts.paypal-context")
```
philwills, line 48: I'm not familiar with using an actorSystem like this.

philwills (approved):

app/services/PaymentServices.scala
alexduf: The reason why we're not using the standard execution context is because paypal's api calls are blocking.
I used what play's [documentation recommend](https://www.playframework.com/documentation/2.5.x/ThreadPools#Using-other-thread-pools)

> In certain circumstances, you may wish to dispatch work to other thread pools. This may include CPU heavy work, or IO work, such as database access. To do this, you should first create a ThreadPool

In this instance I moved the wiring of that execution context out of the wiring package to put it as close as possible from the actual paypal service. I'm actually considering putting it within the service

app/models/PaymentMode.scala
alexduf: It looks nice! I think we have a couple other enums in our codebase, do you mind if migrate them all at once in a separate PR?

app/services/PaymentServices.scala
philwills: Ah, I've just always manually created a separate `ExecutionContext` in this kind of case, as I'm allergic to config files.

app/models/PaymentMode.scala
philwills: Absolutely, I just thought it was worth sharing. No reason why it should block this PR.

PR #173, opened 2016-10-10
Remove recurring payment test

desbo:


alexduf (approved):
Was there no CSS to remove? if not, then all good 👍

desbo:
@alexduf I've kept one file, `notifications.scss`, because it may be useful in the future if we want to show messages inside the form. I removed the import from the main stylesheet so it's not actually being served to users for now though.

PR #174, opened 2016-10-11
Support for automated emails

alexduf:
This is the second part of the work as the first part was merged here https://github.com/guardian/membership-workflow/pull/63

This should not be merged until approved by Rachel.

I know @Mullefa was curious about this PR. @AWare if you want to have a look as well


philwills (approved):
I wonder whether ContributorRow is really pulling it's weight and something like `NotificationSerialisable` that `StripeHook` and `PaypalHook` support might be good, but this is absolutely fine as is and I'd definitely err on the side of merge now and maybe mull that for later.

PR #175, opened 2016-10-12
Simplify running the dev server(s)

alexduf:
I suggest to run play on the port 9112 and webpack on the port 9111
I updated the config, build and documentation accordingly.

I believe it should make things easier, let me know what you think about it.

@desbo @philwills @Mullefa @jranks123

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,15 +35,17 @@ npm install
 npm run devSetup
 ```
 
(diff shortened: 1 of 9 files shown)


desbo (approved):
nice!

philwills:
I need to better understand what grunt is doing and what webpack is doing, but this looks really helpful.

PR #176, opened 2016-10-12
use proper enums

alexduf:
following @philwills's suggestion I used enumeratum to replace all our enums


philwills (approved):
Excellent, that looks a little tidier.

PR #178, opened 2016-10-14
Store Ophan browser ID as well as pageview ID

desbo:
We need to start storing browser IDs as well as pageview IDs (which are currently stored in the `ophanid` column in `contribution_metadata`). These are the changes required to the code to add them to the metadata, there's also a bit to do at the database side:
1. create new columns in the `contribution_metadata` table: `ophan_pageview_id` and `ophan_browser_id`
2. update the `all_payments` view to coalesce `metadata.ophanid` and `metadata.ophan_pageview_id` 
3. release this code, which will start writing IDs to the two new columns
4. move all of the old values from `ophanid` to `ophan_pageview_id`
5. delete the `ophanid` column

@alexduf does this look alright?

### DIFF
--- a/app/controllers/Giraffe.scala
+++ b/app/controllers/Giraffe.scala
@@ -54,7 +54,8 @@ class Giraffe(paymentServices: PaymentServices, addToken: CSRFAddToken) extends 
                           marketing: Boolean,
                           postcode: Option[String],
                           abTests: Set[JsonAbTest],
-                          ophanId: String,
+                          ophanPageviewId: String,
+                          ophanBrowserId: String,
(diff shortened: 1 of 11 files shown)


alexduf:
Just to be thorough I would also update the "ophanId" property we send to stripe in the `pay` method of `Giraffe` (we need to rename that controller one day)

alexduf (changes requested):
It all looks good, aside from my aforementioned comment 👍

alexduf (approved):

alexduf (approved):

PR #180, opened 2016-10-17
Fall cleanup

alexduf:
- remove dead code
- move the stripe payment away from the controller who handles pages


philwills (approved):
I'm especially pleased to see some effort to cut that over-sized giraffe down to size.

🔪

PR #185, opened 2016-10-19
Recurring aa test

Mullefa:
@desbo slight modification to allow for finer grain control on setting complete flags for tests.

### DIFF
--- a/app/views/support/ABTest.scala
+++ b/app/views/support/ABTest.scala
@@ -1,15 +1,10 @@
 package views.support
 
-import java.net.URLEncoder
-import java.nio.charset.StandardCharsets
-
 import com.gu.i18n.CountryGroup
 import com.gu.i18n.CountryGroup._
 import play.api.libs.json._
 import play.api.libs.functional.syntax._
 import play.api.mvc.{Cookie, Request}
-
-
 import scala.util.Random
 import scalaz.NonEmptyList
 import views.support.AmountHighlightTest.AmountVariantData
@@ -120,7 +115,7 @@ object Test {
   val CookiePrefix     = "gu.contributions.test"
   val TestIdCookieName = s"$CookiePrefix.id"
 
-  val allTests = List(AmountHighlightTest, ReducedCheckoutTest)
+  val allTests = List(AmountHighlightTest, ReducedCheckoutTest, AARecurringTest)
 
   def cookieName(v: Variant) = s"$CookiePrefix.${v.testSlug}"
   def cookieName(t: TestTrait) = s"$CookiePrefix.${t.slug}"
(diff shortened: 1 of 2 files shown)


assets/javascripts/src/modules/abTests.es6
```diff
@@ -8,11 +8,9 @@ import { SET_AMOUNT } from 'src/actions';
 export function init() {
     const state = store.getState();
 
-    ophan.loaded.then(function (ophan) {
-        ophan.record({
-            abTestRegister: abTestData(state.data.abTests, false)
-        })
-    });
+    registerTestsWithOphan(state.data.abTests, initTestFlags);
```
desbo: I'm not sure this really needs the second parameter. At this point it's fine to just send `complete: false` for all of the active tests. Sending `complete: true` is down to each individual test, and might happen at any point in the app's lifecycle so it should be enough to just provide a helper function that fires the `true` event for a test.

assets/javascripts/src/modules/abTests.es6
```diff
@@ -21,27 +19,94 @@ export function init() {
...
+   return isAARecurringTest(test);
+}
+
+/**
+ * To be used with registerTestsWithOphan().
+ * Set a test's complete flag to true, apart from the AA recurring test which is ignored.
+ */
+function completeTestFlags(test) {
+    if (isAARecurringTest(test)) {
+        return null;
```
desbo: My idea for this test was to fire `true` straight away so we should see as many `falses` as there are `true`s

Mullefa:
@desbo based on your comments I've removed the second parameter from `registerTestsWithOphan()`. Now, when the page loads, all test complete flags are set to false. I realised for the AA test this is enough, so I haven't bothered setting its flag to true at any point.

Regarding the helper function for setting a test complete flag to true, I thought it could be implemented as needed, but happy to do it as part of this pull request if you want?

desbo:
> Now, when the page loads, all test complete flags are set to false. I realised for the AA test this is enough, so I haven't bothered setting its flag to true at any point.

The problem before was that we weren't seeing as many `true`s as there were events in GA. I would still have the A/A test fire a `true` event as well as `false`, but for this test it's fine to fire `true` as soon as the page loads. 

> Regarding the helper function for setting a test complete flag to true, I thought it could be implemented as needed, but happy to do it as part of this pull request if you want?

Because we only run one test a time, the code would be exactly the same as `registerTestsWithOphan` but with `complete` set to `true`. I would just update that function to take a boolean parameter for `complete`.  You then run that function with `false` at the start to register everything and  the test can call it with `true` later.

app/views/support/ABTest.scala
```diff
@@ -162,6 +157,18 @@ object Test {
   }
 }
 
+/**
+ * The first time the recurring test was run, there was a big discrepancy between the data in Ophan and GA.
+ * This test fires an event to Ophan on page load; (independently) an event will also be fired to GA on page load.
+ * By comparing the event counts, we hope to debug the discrepancy.
+ */
```
philwills, line 164: 👍 for comments that convey rationale rather than implementation

desbo (approved):

PR #186, opened 2016-10-20
Contrib cookie followup

Mullefa:
@guardian/contributions 

Fixed bugs raised my @desbo - the cookie set now has a max age of 10 years, and if on production, has domain `.theguardian.com`.

Whilst doing this, I realised the pattern used for allowing the syntax `result.setContribTimestampCookie()`could be extended / simplified to work for all cookies. I like it, as it makes it very explicit which cookies are currently being used by the application, and what can types can be used to set their value.


app/configuration/Config.scala
```diff
@@ -22,4 +23,5 @@ object Config {
     (idWebAppUrl / "signin") ? ("returnUrl" -> s"$contributeUrl$uri") & idSkipConfirmation
   private val idSkipConfirmation: (String, String) = "skipConfirmation" -> "true"
 
+  val guardianDomain = ".theguardian.com"
```
desbo, line 26: I'm not sure this is necessary, won't the host from config value above already be set to theguardian.com in prod?

Mullefa: Are you referring to `Config.domain`? This is derived from `contribute.url` which on production is set to `contribute.theguardian.com`, so I don't think we could use that.

app/cookies/ContribTimestampCookie.scala
```diff
@@ -0,0 +1,38 @@
...
+import java.time.Instant
+
+import com.gu.stripe.Stripe.Charge
+import com.paypal.api.payments.Payment
+import configuration.Config
+
+/**
+ * Should be set when a user contributes to the Guardian.
+ */
+trait ContribTimestampCookie extends CookieType {
```
jranks123: Probably should use full names i.e ContributionsTimestampCookie

app/cookies/syntax/package.scala
```diff
@@ -0,0 +1,21 @@
+package cookies
+
+import play.api.mvc.Result
+
+package object syntax {
+
+  /**
+   * Allows for the following syntax: `result.setCookie[C].using(data)`
+   */
+  implicit class CookieRequest(result: Result) {
```
alexduf: this needs to extends `AnyVal` to avoid any runtime over-head

app/cookies/CookieType.scala
```diff
@@ -0,0 +1,40 @@
+package cookies
+
+import play.api.mvc.Cookie
+
+abstract class CookieType {
```
alexduf: as we talked about, I got confused by the name of `CookieType`, not realising it was the definition of a cookie rather than containing the value of the cookie. I think it could be renamed to something more expllicit, including its sub-class `ContribTimestampCookie`

app/configuration/Config.scala
alexduf: I see no problem with an extra config attribute here, especially as we're dealing with two separate concepts: the main website and the contribution page

app/cookies/CookieType.scala
```diff
@@ -0,0 +1,40 @@
...
+  val path: String = "/"
+  val domain: Option[String] = None
+  val secure: Boolean = false
+  val httpOnly: Boolean = false
+}
+
+/**
+ * Represents data of type [[A]] being able to set the value for a given cookie type [[C]].
+ */
+trait CookieValue[A, C <: CookieType] {
```
alexduf: As we talked about, I think as much as I like that nifty implicit conversion, we may as well directly go for a string at the time we set the cookie. That will simplify the code and make it more straightforward. (less beautiful though 😃  )

app/cookies/CookieType.scala
```diff
@@ -0,0 +1,40 @@
...
+  def value(data: A): String
+}
+
+object CookieType {
+
+  /**
+   * Create a cookie from a cookie type, using the data to set its value.
+   */
+  def createCookie[C <: CookieType, A](data: A)(implicit c: C, cv: CookieValue[A, C]): Cookie = {
```
alexduf: I think the variable `c` deserves a full name

alexduf (changes requested):

alexduf (approved):

alexduf:
Nice 👍 ⭐️

PR #187, opened 2016-10-25
Don't send automatic email for some CMP

alexduf:
When a user comes back to the contribution website we don't want to send the same automatic email that was sent the first time.
This PR includes a change that moves when the email is sent in the workflow.
It is now sent when we save the metadata as it happen when: we know the user has paid, and we know the user's CMP code.

### DIFF
--- a/app/services/StripeService.scala
+++ b/app/services/StripeService.scala
@@ -49,7 +49,8 @@ class StripeService(
       created = created,
       amount = amount,
       currency = currency,
-      name = name
+      name = name,
+      cmp = cmp
     ))
 
     val metadata = ContributionMetaData(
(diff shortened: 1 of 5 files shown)


alexduf:
@guardian/contributions

app/services/StripeService.scala
```diff
@@ -35,13 +35,23 @@ class StripeService(
...
+    idUser: Option[IdentityId],
+    amount: BigDecimal,
+    currency: String
   ): XorT[Future, String, SavedContributionData] = {
 
     // Fire and forget: we don't want to stop the user flow
     idUser.map { id =>
       identityService.updateMarketingPreferences(id, marketing)
     }
+    emailService.thank(ContributorRow(
```
philwills, line 47: It seems a bit strange to do this only for effect, discarding the XorT for Stripe, but to include it in the result for Paypal.

philwills (approved):
Minor quibble aside, this looks good.

app/services/StripeService.scala
alexduf: You're entirely right, to be honest I put it there in the stripe service because it was a convenient way to order the tasks.

PR #189, opened 2016-10-25
Don't parse "name" as we don't use it anymore

alexduf:
Following the two steps tests, our hooks weren't containing the `name` field in their metadata. We didn't catch that it was causing the stripe hook to fail.

Following this morning's PR where I moved the email to be sent earlier, we don't need to parse the name anymore.

This fix remove the parsing of the name and therefore fixes the webhook processing


alexduf:
@guardian/contributions

philwills (approved):

PR #190, opened 2016-10-25
Implemented autofill

jranks123:
Since we moved to React, the autofilling of signed-in users email addresses and names from their Identity details has not been working. This PR re-implements this functionailty.


alexduf (approved):
nice!

PR #191, opened 2016-10-25
debugging GA events not firing

Mullefa:
@guardian/contributions 

This pull request fixes a bug which prevents events being fired for the `AARecurringTest`. The bug was due to trying to fire a GA event before GA had initialised.

As suggested, to fix this I have created a function which returns a promise which resolves only when GA has been initialised (i.e. `window.ga` is not `undefined`).

I have deployed the branch to `CODE` and [event tracking tracker](https://chrome.google.com/webstore/detail/event-tracking-tracker/npjkfahkbgoagkfpkidpjdemjjmmbcim) reports that the event is now fired.


alexduf (approved):
👍

PR #192, opened 2016-10-26
Let contributions frontend read the new SSH key bucket

philwills:
cc @guardian/contributions


desbo (approved):

philwills:
I don't think this is actually necessary in retrospect. I believe we can just remove these old perms as there's a separate new one.

PR #194, opened 2016-10-27
Correctly handle refunds in Stripe webhooks

desbo:
Before this change, we tried to detect refunds by reading the "status"
property of a webhook as "refunded". Stripe doesn't use that property
for refunds, instead there's a separate "refunded" boolean flag on each
charge, which we now use to determine the status in the DB.

@philwills


philwills (approved):
I'm going to merge this now, but it would be nice to have a test with some real JSON minus any PI data to demonstrate the behaviour.

PR #196, opened 2016-11-01
fetch the country code for paypal payments

alexduf:
When the webhook comes back, query the paypal API to fetch the country code.
Both paypal and stripe use the 2 characters ISO 3166-1 country code, so I'm putting them in the same column.
The column is named cardCountry and does not exactly represent what it now contains, I'm not sure if it is worth renaming (opinion appreciated on the matter)

@guardian/contributions

### DIFF
--- a/app/models/PaymentHook.scala
+++ b/app/models/PaymentHook.scala
@@ -63,13 +63,13 @@ case class PaymentHook(
 )
 
 object PaymentHook {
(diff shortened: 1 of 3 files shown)


philwills (approved):
This looks sensible.

Is the intention here that if we can't get the country, we should fail the whole hook and rely on Paypal's retry mechanism?

With regards to the question of the column name, I'd prefer it to change, but think that's a separate and less pressing PR.

PR #197, opened 2016-11-01
Update to latest membership-common

paulbrown1982:
The latest Membership-common separated out Ireland as its own country group, rather than it being part of Europe. This was so that the commercial team could have a specific European country to reference on the checkout page - a selected country is required by Subs and Membership promotion logic. This change updates Contributions to handle requests from Ireland like it always has done.

cc @philwills @desbo


desbo (approved):
LGTM

PR #200, opened 2016-11-07
Allow us to use any EC2 instance type

philwills:
I'd rather not have to update CFN if for some reason we want to use beefier instances.


philwills:
cc @guardian/contributions

alexduf (approved):
👍

PR #201, opened 2016-11-07
[!!!!!!!!!DO NOT MERGE!!!!!!!!] Minimal change to disable Paypal

philwills:
In case of emergency, remove button

This is meant as a discussion point for the smallest change if there were a problem with one of our payment providers, so that people don't have to worry about it in the heat of the moment.


philwills:
cc @guardian/contributions

jranks123 (approved):

PR #213, opened 2016-11-15
Collect country code at payment time

alexduf:
Get the country code at the time of the payment rather than the time of the webhook

This is following: https://github.com/guardian/contributions-platform/pull/31

It saves the country at the time of the payment, both for paypal and stripe.

This is to avoid potential trouble if paypal were to hit our webhook endpoint a bit too much again.

@guardian/contributions


philwills (approved):
I quite liked the explicitness of `cardCountry` as a field, but otherwise 👍

alexduf:
Me too but `cardCountry` is a concept that only makes sense for stripe as paypal's field is coming from the user's information. So I used the biggest common denominator.

PR #214, opened 2016-11-16
Fix twitter metadata

desbo:
This adds a new image for Twitter cards, and a couple of extra bits of metadata. I've also changed the Twitter username to @guardian since I think that suits us better than membership. I think the old values were just copied over from the membership repo.


alexduf (approved):
Nice!

PR #218, opened 2016-11-18
Simplify Cloudformation with YAML features

philwills:
I've successfully deployed this to CODE.


alexduf (approved):
This looks beautiful, especially the user data 👍 

Why did you remove the github keys integration ? was it not working?

philwills:
@alexduf it was actually duplicated elsewhere in the template. That should still work.

PR #220, opened 2016-11-24
Upgrade to cats 0.8.1

philwills:


alexduf (changes requested):
Looks good, but it looks like TeamCity is grumpy

philwills:
@alexduf I saw it lurking in my working directory, so pushed it, but apparently forgot that I hadn't actually got it to compile.

philwills:
Now updated so that it actually compiles.

alexduf (approved):
👍

PR #224, opened 2016-12-21
Cloudwatch logs & riff-raff.yaml

alexduf:
Let's use bleeding edge technology!
 - cloudwatch log agent is now baked in the AMI
 - nicer yaml / cloudformation syntax
 - deploy.json is so passé


conf/riff-raff.yaml
```diff
@@ -0,0 +1,21 @@
+stacks: [membership]
+regions: [eu-west-1]
+deployments:
+  cfn:
+    type: cloud-formation
+    app: contributions-frontend
+    parameters:
+      prependStackToCloudFormationStackName: false
```
philwills, line 8: Might be worth trying to switch to `cloudFormationStackByTags` from https://riffraff.gutools.co.uk/docs/magenta-lib/types#cloudformation which is the default for riff-raff.yml, then we don't need the prepend nonsense.

conf/riff-raff.yaml
```diff
@@ -0,0 +1,21 @@
...
+      cloudFormationStackName: contributions-frontend
+      templatePath: cfn.yaml
+  contributions-frontend-ami:
+    type: ami-cloudformation-parameter
+    app: contributions-frontend
```
philwills, line 13: It's interesting how often this is repeated, feels like a bit of a smell cc @sihil

philwills (approved):
This looks awesome.

conf/riff-raff.yaml
alexduf: I believe it's already searching by tags as `true` is the default value for `cloudFormationStackByTags`

I kept `prependStackToCloudFormationStackName` just in case we would destroy the stack and let riff-raff re-create it, to keep a similar name

philwills: I'm pretty sure you can only update, not create a CloudFormation stack with RiffRaff.

alexduf:
Don't you worry @prout-bot I'm on it :)

PR #225, opened 2017-01-09
Autofill real name

alexduf:
Only autofill when we have both the first name and the last name


alexduf:
cc @guardian/contributions

assets/javascripts/src/modules/contribute.es6
```diff
@@ -35,20 +35,19 @@ export function init() {
 
 
 function autoFill() {
-    let IDENTITY_API = 'https://idapi.theguardian.com/user/me/';
+    let IDENTITY_API = '/user/autofill';
```
philwills: Seems confusing to continue to call this `IDENTITY_API`

philwills (approved):
👍

PR #226, opened 2017-01-24
New stripe

AWare:
This is a first pass at the new stripe checkout. There are a few more todos:

- [ ] Whether this needs an AB test (and how)
- [ ] Removal of old stripe code
- [ ] Mobile

What I'm looking for is general approval of the way I've gone about doing it, and the refactorings I've made. If you're all happy with this, I'll progress with it and have a real pull request out ASAP.

@desbo @dominickendrick @jranks123 @alexduf 

image
image

### DIFF
--- a/app/views/fragments/javaScriptFirstSteps.scala.html
+++ b/app/views/fragments/javaScriptFirstSteps.scala.html
@@ -7,6 +7,7 @@
 
     @* Critical scripts that should execute as soon as possible, preventing layout thrashing *@
 <script type="text/javascript" src="https://js.stripe.com/v2/"></script>
+<script src="https://checkout.stripe.com/checkout.js"></script>
 
         <!--[if IE 9]>
     <script src="@Asset.at("javascripts/lib/polyfills.min.js")"></script>
@@ -27,7 +28,15 @@
         },
         @for(stripePublicKey <- pageInfo.stripePublicKey) {
         stripePublicKey: '@stripePublicKey',
-        }
+            stripeCheckout : {
+             key: '@stripePublicKey',
+             image: 'https://d24w1tjgih0o9s.cloudfront.net/gu.png',
+             locale: 'auto',
+             name: 'Guardian',
+             zipCode: false,
+             allowRememberMe: false
+         },
+            }
         membership: {
             buildNumber: '@app.BuildInfo.buildNumber',
             svgSprite: "@Asset.at("images/svg-sprite.svg")"
--- a/assets/javascripts/src/actions.es6
+++ b/assets/javascripts/src/actions.es6
@@ -39,9 +39,8 @@ export function openStripeCheckout(dispatch) {
         dispatch({type:CLOSE_STRIPE});
     };
     let process = (token) => {
+        dispatch({ type: SUBMIT_PAYMENT });
         let data = paymentFormData(state, token.id);
-        console.log('testy');
-        console.log(state);
         fetch(urls.pay, {
             credentials: 'same-origin',
             method: 'POST',
--- a/assets/javascripts/src/actions.es6
+++ b/assets/javascripts/src/actions.es6
@@ -5,6 +5,8 @@ import { urls } from 'src/constants';
 import * as stripe from 'src/modules/stripe';
 import { trackCheckout, trackPayment } from 'src/modules/analytics/ga';
 
+import {showCheckout} from 'src/modules/stripeCheckout'
+
 export const SET_DATA = "SET_DATA";
 export const SET_COUNTRY_GROUP = "SET_COUNTRY_GROUP";
 
@@ -14,6 +16,8 @@ export const SET_AMOUNT = "SET_AMOUNT";
 export const UPDATE_DETAILS = "UPDATE_DETAILS";
 export const UPDATE_CARD = "UPDATE_CARD";
 
+export const OPEN_STRIPE = "OPEN_STRIPE";
+export const CLOSE_STRIPE = "CLOSE_STRIPE";
 export const SUBMIT_PAYMENT = "SUBMIT_PAYMENT";
 export const PAYMENT_COMPLETE = "PAYMENT_COMPLETE";
 export const PAYMENT_ERROR = "PAYMENT_ERROR";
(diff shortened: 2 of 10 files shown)


app/views/fragments/javaScriptFirstSteps.scala.html
```diff
@@ -27,7 +28,15 @@
         },
         @for(stripePublicKey <- pageInfo.stripePublicKey) {
         stripePublicKey: '@stripePublicKey',
-        }
+            stripeCheckout : {
+             key: '@stripePublicKey',
+             image: 'https://d24w1tjgih0o9s.cloudfront.net/gu.png',
```
alexduf, line 33: I would prefer having that asset hosted within the repo (if that's possible?)

alexduf (commented):
Aside from the obvious `console.log` (which are okay for a WIP) it looks like it's going in the right direction, thanks Alex!

app/views/fragments/javaScriptFirstSteps.scala.html
AWare: Possible, but this asset is common between the three implementations so far.

assets/javascripts/src/actions.es6
```diff
@@ -28,6 +32,47 @@ export const GA_ENABLED = "GA_ENABLED";
 
 export const AUTOFILL = "AUTOFILL";
 
+export function openStripeCheckout(dispatch) {
+    const state = store.getState();
+    console.log('is this thing on');
```
paulbrown1982, line 37: Don't forget to remove this before merging

assets/javascripts/src/actions.es6
```diff
@@ -28,6 +32,47 @@ export const GA_ENABLED = "GA_ENABLED";
...
+            method: 'POST',
+            headers: { 'Content-Type': 'application/json' },
+            body: JSON.stringify(data)
+        })
+        .then(response => response.json() // response: Response
+            .then(json => {
+                return { response: response, json: json }
+            }))
+            .then((response) => {
+                if (response.response.ok)
```
paulbrown1982, line 55: Not that any of this is incorrect, but I'm a little confused about exactly which variable is the actual response ...

assets/javascripts/src/components/Main.jsx
```diff
@@ -45,6 +49,7 @@ function mapStateToProps(state) {
 }
 
 function mapDispatchToProps(dispatch) {
+    console.log('hey',openStripeCheckout);
```
paulbrown1982, line 52: Another bit of debugging to remember to remove.

assets/javascripts/src/components/Main.jsx
```diff
@@ -58,10 +63,16 @@ function mapDispatchToProps(dispatch) {
             dispatch({ type: UPDATE_DETAILS, details: d })
         },
         updateCard: c => dispatch({ type: UPDATE_CARD, card: c }),
+        openStripe: () => {dispatch()},
+        //closeStripe: () => {dispatch({type: CLOSE_STRIPE})},
```
paulbrown1982, line 67: Why is this commented out?

assets/javascripts/src/actions.es6
```diff
@@ -28,6 +32,47 @@ export const GA_ENABLED = "GA_ENABLED";
 
 export const AUTOFILL = "AUTOFILL";
 
+export function openStripeCheckout(dispatch) {
+    const state = store.getState();
+    console.log('is this thing on');
+    let hide = () => {
```
desbo, line 38: not sure having this here helps readability, probably just declare the lambda in the `showCheckout` call

assets/javascripts/src/actions.es6
```diff
@@ -28,6 +32,47 @@ export const GA_ENABLED = "GA_ENABLED";
...
+        dispatch({type:CLOSE_STRIPE});
+    };
+    let process = (token) => {
```
desbo, line 41: should be a `const` rather than `let`. 
i'd probably define it inside the stripe module and import it rather than inside this function

assets/javascripts/src/actions.es6
desbo: it does get a little confusing but i think the idea is to persist the original `response` object along with the parsed json... it might be worth writing a `fetch-json` module that does this

assets/javascripts/src/components/Navigation.jsx
```diff
@@ -21,17 +21,42 @@ export default class Navigation extends React.Component {
         const isFirstPage = !processing && this.props.page === PAGES.CONTRIBUTION;
         const showMobileBack = !processing && this.props.page == PAGES.PAYMENT;
         const showBack = !processing && this.props.page !== PAGES.CONTRIBUTION;
-        const showPay = !processing && !!this.props.amount && this.props.page === PAGES.PAYMENT;
-        const showForward = !processing && this.props.page == PAGES.DETAILS;
+        const showPay = !processing && !!this.props.amount && this.props.page === (this.props.stripeCheckout?PAGES.DETAILS:PAGES.PAYMENT);
+        const showNext = (!this.props.stripeCheckout) && !processing && this.props.page == PAGES.DETAILS;
+        const classes = {
+            nav: () => {'contribute-navigation ' + this.classNameFor(this.props.page)},
```
desbo, line 27: does this need to be a function?

assets/javascripts/src/components/Navigation.jsx
```diff
@@ -21,17 +21,42 @@ export default class Navigation extends React.Component {
         const isFirstPage = !processing && this.props.page === PAGES.CONTRIBUTION;
         const showMobileBack = !processing && this.props.page == PAGES.PAYMENT;
         const showBack = !processing && this.props.page !== PAGES.CONTRIBUTION;
-        const showPay = !processing && !!this.props.amount && this.props.page === PAGES.PAYMENT;
-        const showForward = !processing && this.props.page == PAGES.DETAILS;
+        const showPay = !processing && !!this.props.amount && this.props.page === (this.props.stripeCheckout?PAGES.DETAILS:PAGES.PAYMENT);
+        const showNext = (!this.props.stripeCheckout) && !processing && this.props.page == PAGES.DETAILS;
+        const classes = {
```
desbo, line 26: not sure defining the classes in the object and then referencing them in the tag makes things any clearer

desbo:
A couple of tweaks to make but looks like a good enough start. I've merged in a change to the A/B testing that allows us to limit the audience size for the Stripe test. You'll need to check for the `stripe-checkout` test slug in the `data-ab-tests` attribute (which I'm sure gets added to the app state at some point) and show either the original or new checkout based on the variant (either `control` or `stripe`).

desbo:
OK, ignore my last comment. I'm going to close this PR for now and will work on incorporating the checkout with the new A/B testing code.

PR #228, opened 2017-02-01
Add simplified A/B test allocation tool

desbo:
This is a rewrite of the existing `TestTrait` setup that gives us control over audience segmentation by setting a size and offset on your test (just like in frontend). It'll allow us to measure the impact of the new Stripe checkout flow against a smallish percentage of the audience.

Apart from the addition of segmentation, it's quite a bit simpler than the version before (which was copied over from the membership project), removing the ability for variants within the same test to have separate weights and country targeting. These shouldn't be too difficult to add if we need them again.

I've added a dummy Stripe test to the new `Test` object, the frontend work will be in a separate PR.

@guardian/contributions


app/actions/CommonActions.scala
```diff
@@ -27,4 +29,20 @@ object CommonActions {
       }
     }
   }
+
+  case class ABTestRequest[A](testId: Int, request: Request[A]) extends WrappedRequest(request) {
```
alexduf: Am I right to assume `testId` is a user related id rather than a test related id?

alexduf, line 33: Am I right to assume `testId` is a user related id rather than a test related id?

app/abtests/Test.scala
```diff
@@ -0,0 +1,54 @@
+package abtests
+
+import com.github.slugify._
+import play.api.libs.json.{JsValue, Json, Writes}
+import play.api.mvc.Cookie
+
+case class Test(name: String, audienceSize: Double, audienceOffset: Double, variants: Variant*) {
```
alexduf: given that the id generated is an int, would that make sense to have these as ints, and avoid the math below that may introduce rounding errors?

alexduf (approved):
Looks good, two nitpicks, but nothing to worry about 👍

app/actions/CommonActions.scala
desbo: yes you are. I'll give it a better name

app/abtests/Test.scala
desbo: I think it only makes sense if the maximum test ID is 100, in which case you can think of an `Int` size and offset as a percentage. Otherwise it probably makes more sense (or is at least more general) to have them as `Doubles`.

alexduf: cool

Mullefa: There's a few cases which the `Test` logic doesn't cover e.g. 
```
val idRange = 0 to 5
val variants = ??? // List with 4 variants 
// omits two values in the id range
idRange.grouped(idRange.length / variants.size)
  .zip(variants.iterator)
```
(I'm guessing this isn't the desired behaviour?)

PR #229, opened 2017-02-03
Add support for apps:

alexduf:
- apps will send us a special get parameter that is recognised and stored in session
 - when a payment is processed, the server redirects app users back into the app using a deep-link

cc @guardian/contributions


app/controllers/Redirect.scala
```diff
@@ -8,4 +10,8 @@ trait Redirect {
     val queryParamsToForward = Set("INTCMP", "CMP", "REFPVID") ++ additionalParams
     Redirect(destinationUrl, request.queryString.filterKeys(queryParamsToForward), SEE_OTHER)
   }
+
+  def thankYouMobileUri(amount: ContributionAmount): String = {
+    s"x-gu://contribution?date${LocalDate.now().toString}&amount=$amount"
```
desbo, line 15: out of interest, what's the amount used for in the app?

desbo (approved):
good stuff

app/controllers/Redirect.scala
alexduf: it's stored and sent back to olgil, so we can decide what to do. For instance if someone pays £1,000 we can decide they get no creative for 10 years

Realisticly we won't use it in the near future, just the date

alexduf:
Do you mind having another look @desbo ? the important bit is here and shows how I worked around the conflict we had between our two PRs: https://github.com/guardian/contributions-frontend/pull/229/commits/73b6929a7f11ec4c182ff7cf5ba9c2ef0237bb1a

Spoiler, I only solved the problem for the mobile side of things, couldn't apply the same magic on the ab test side.

desbo (approved):

PR #235, opened 2017-02-23
App flow optimisation

alexduf:
This will change:

 - android app users paying with paypal: will see the email page before being redirected
 - ios app users paying with paypal or stripe: will be redirected to the thank you page, onto which a button will be there

The idea behind that change in the worflow is to warn the ios app user about being redirected as otherwise they will see that odd popup asking them if they want to open the guardian app.

cc @guardian/contributions 

iphone5

### DIFF
--- a/app/controllers/Contributions.scala
+++ b/app/controllers/Contributions.scala
@@ -7,6 +7,7 @@ import com.gu.i18n.CountryGroup._
 import com.gu.i18n._
 import com.netaporter.uri.dsl._
 import configuration.Config
+import models.ContributionAmount
 import play.api.mvc._
 import play.filters.csrf.{CSRF, CSRFAddToken}
 import services.PaymentServices
@@ -89,12 +90,17 @@ class Contributions(paymentServices: PaymentServices, addToken: CSRFAddToken) ex
     val charge = request.session.get("charge_id")
     val title = "Thank you!"
 
+    val iosRedirectUrl = request.session.get("amount")
+      .flatMap(ContributionAmount.apply)
+      .map(mobileRedirectUrl)
+      .filter(_ => request.isIos)
+
     Ok(views.html.giraffe.thankyou(PageInfo(
       title = title,
       url = request.path,
       image = None,
       description = Some("You’ve made a vital contribution that will help us maintain our independent, investigative journalism")
-    ), social, countryGroup, charge))
+    ), social, countryGroup, charge, iosRedirectUrl))
   }
 }
 
--- a/app/controllers/PaypalController.scala
+++ b/app/controllers/PaypalController.scala
@@ -3,6 +3,7 @@ package controllers
 import actions.CommonActions._
 import cats.data.EitherT
 import cats.instances.future._
+import cats.syntax.show._
 import cookies.ContribTimestampCookieAttributes
 import cookies.syntax._
 import com.gu.i18n.{CountryGroup, Currency}
@@ -71,7 +72,7 @@ class PaypalController(ws: WSClient, paymentServices: PaymentServices, checkToke
 
       val session = List(
         "email" -> savedData.contributor.email
-      ) ++ amount.map("amount" -> _.toString)
+      ) ++ amount.map("amount" -> _.show)
 
       redirectWithCampaignCodes(redirectUrl)
         .addingToSession(session :_ *)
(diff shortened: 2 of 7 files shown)


app/controllers/Contributions.scala
```diff
@@ -89,12 +90,17 @@ class Contributions(paymentServices: PaymentServices, addToken: CSRFAddToken) ex
     val charge = request.session.get("charge_id")
     val title = "Thank you!"
 
+    val iosRedirectUrl = request.session.get("amount")
```
Mullefa, line 93: My personal preference would be to use a `for` comprehension. Easy either way though.

app/controllers/PaypalController.scala
```diff
@@ -190,17 +192,22 @@ class PaypalController(ws: WSClient, paymentServices: PaymentServices, checkToke
...
+  def updateMetadata(countryGroup: CountryGroup) = NoCacheAction.async(parse.form(metadataUpdateForm)) { implicit request =>
+    val paypalService = paymentServices.paypalServiceFor(request)
+    val marketingOptIn = request.body.marketingOptIn
+    val idUser = IdentityId.fromRequest(request)
+    val contributor = request.session.data.get("email") match {
+      case Some(email) => paypalService.updateMarketingOptIn(email, marketingOptIn, idUser).value
+      case None => Future.successful(Logger.error("email not found in session while trying to update marketing opt in"))
+    }
+
+    val url = request.session.get("amount").flatMap(ContributionAmount.apply)
```
Mullefa, line 204: My personal preference would to be use a `for` comprehension. Easy either way though.

app/controllers/PaypalController.scala
```diff
@@ -190,17 +192,22 @@ class PaypalController(ws: WSClient, paymentServices: PaymentServices, checkToke
...
-        case None => Future.successful(Logger.error("email not found in session while trying to update marketing opt in"))
-      }
-      contributor.map { _ =>
-        Redirect(routes.Contributions.thanks(countryGroup).url, SEE_OTHER)
-      }
+  def updateMetadata(countryGroup: CountryGroup) = NoCacheAction.async(parse.form(metadataUpdateForm)) { implicit request =>
+    val paypalService = paymentServices.paypalServiceFor(request)
+    val marketingOptIn = request.body.marketingOptIn
+    val idUser = IdentityId.fromRequest(request)
+    val contributor = request.session.data.get("email") match {
```
Mullefa, line 199: Another approach without having to extract the `.value` of the update result (which seems like an anti-pattern to me):

```
val email =  request.session.data.get("email")

EitherT.fromOption[Future](email, "email not found in session while trying to update marketing opt in")
  .flatMap(paypalService.updateMarketingOptIn(_, marketingOptIn, idUser))
  .fold(message => Logger.error(message), _ => ())
  .map(_ => Redirect(url, SEE_OTHER))
```

Mullefa (approved):
A couple of comments on the code, but probably just comes down to personal preference - therefore, feel free to ignore. 

Also, whilst reviewing this PR I had some insightful and helpful chats with @alexduf about the different user flows. This made reviewing the pull request a lot easier. With this in mind, could we include this information in the README? If it's outside the scope of this pull request, I'd be happy to update the documentation in a separate PR.

app/controllers/PaypalController.scala
alexduf: I take good note of this one, but it's a bit out of scope as I really just touched the formatting here. I'll create another PR to surface your feedback

app/models/ContributionAmount.scala
```diff
@@ -20,4 +19,6 @@ object ContributionAmount {
       currency <- Currency.fromString(currencyString)
     } yield ContributionAmount(number, currency)
   }
+
+  implicit val showContributionAmount: Show[ContributionAmount] = Show.show(f => f"${f.amount}%1.2f${f.currency}")
```
TBonnin: Not necessary related to this PR but currency formatting is a topic by itself (and quite complicated as well)
Have we considered internationalisation of the format? Some country prefers to put the currency sign/name before, thousands and cents delimiters might be different, etc...

As an european I am always disturbed when americans put the `euro` sign before (€9.99 😱 ). I guess it would be the same the other way around.

alexduf: Luckily this is a string that isn't displayed, or isn't planned to be displayed to the user. It is stored on the app, then sent to our campaign management tool to avoid showing membership campaigns to paying users.

So I went for the technical formatting:
50.0GBP or 25.0EUR

It's meant to be easily parseable just in case we want to treat someone who gives 2000 and 20 differently, but nothing is planned in that direction yet.

TBonnin: Got it.

PR #236, opened 2017-03-03
Stripe Checkout

desbo:
👀 
mar 03 2017 18 07

This also includes an upgrade to webpack 2 and using the NPM ophan package and including it in the bundled js, allowing us to get rid of curl.

@guardian/contributions


jranks123 (approved):
👍

PR #238, opened 2017-03-08
Fix A/B test tracking

desbo:


assets/javascripts/src/modules/abTests.es6
```diff
@@ -41,6 +41,8 @@ function registerTestsWithOphan(tests, complete) {
             'variantName': test.variantSlug,
             'complete': String(complete)
         }
+
+        return obj;
```
desbo, line 45: Note that this was actually the bug, the other stuff was just housekeeping.

alexduf (approved):

PR #241, opened 2017-03-16
Remove header in Android app

maxspencer:
At the moment, when a user visits the contribute page view the app, they see a kind of double header (shown below). In this PR I've tried to use the `platform` query string param we've started sending to remove parts of the template.

This was my first time looking at Twirl templates and I really don't know if this whole approach is right, so I invite robust feedback!

screenshot_20170316-135827

### DIFF
--- a/app/controllers/Contributions.scala
+++ b/app/controllers/Contributions.scala
@@ -57,6 +57,7 @@ class Contributions(paymentServices: PaymentServices, addToken: CSRFAddToken) ex
       val intCmp = request.getQueryString("INTCMP")
       val refererPageviewId = request.getQueryString("REFPVID")
       val refererUrl = request.headers.get("referer")
+      val platform = request.getQueryString("platform")
 
       val pageInfo = PageInfo(
         title = "Support the Guardian | Contribute today",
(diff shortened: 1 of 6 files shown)


app/controllers/Contributions.scala
```diff
@@ -57,6 +57,7 @@ class Contributions(paymentServices: PaymentServices, addToken: CSRFAddToken) ex
       val intCmp = request.getQueryString("INTCMP")
       val refererPageviewId = request.getQueryString("REFPVID")
       val refererUrl = request.headers.get("referer")
+      val platform = request.getQueryString("platform")
```
alexduf, line 60: Ideally I'd like to see a platform agnostic flag to be passed to the template.
Something like `showHeader` would be a good idea.

Also, we have a utility function: `request.isAndroid` would do the trick

alexduf (commented):
I think you also have compilation errors to fix.

app/controllers/Contributions.scala
maxspencer: Those both sound like good improvements, thanks. I'll set the project up properly on my own machine to ensure everything builds, I was just being lazy before, not wanting to get AWS credentials and so on.

PR #242, opened 2017-03-20
Roll out Stripe to 100%

desbo:
This releases the Stripe checkout flow to everyone. Although this performed 3.8% worse than the control, we are releasing it now to avoid any danger of being fined for not meeting PCI compliance, and will prioritise implementing [Stripe Elements](https://stripe.com/docs/elements) (which will allow us to keep the original design and be PCI compliant) in an upcoming sprint. 

@guardian/contributions

### DIFF
--- a/app/abtests/Test.scala
+++ b/app/abtests/Test.scala
@@ -38,8 +38,10 @@ object Test {
   def cmpCheck(pattern: Regex)(r: Request[_]): Boolean =
     r.getQueryString("INTCMP").exists(pattern.findAllIn(_).nonEmpty)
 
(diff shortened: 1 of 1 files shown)


dominickendrick (approved):

desbo:
as discussed @dominickendrick we'll wait until Amanda is back in the office to run this by her before releasing

paulbrown1982:
It would be great to get that 3.8% broken down by screen sizes (or mobile/desktop), as it's the small-screen mobile experience where Stripe Checkout pops up a new tab rather than an overlay.

PR #243, opened 2017-03-22
Refactor Test class to allow for canRun checks

desbo:
In addition to audience segmentation, you can now create tests with or without can run criteria, which I've called "split tests" and "conditional tests". The can run check is a boolean function that takes a request as its only parameter. I figure this should be enough for most use cases.

@guardian/contributions

### DIFF
--- a/app/abtests/Test.scala
+++ b/app/abtests/Test.scala
@@ -2,13 +2,20 @@ package abtests
 
 import com.github.slugify.Slugify
 import play.api.libs.json.{JsValue, Json, Writes}
-import play.api.mvc.Cookie
+import play.api.mvc.{Cookie, Request}
+
+import scala.util.matching.Regex
 
 case class Percentage(value: Double) {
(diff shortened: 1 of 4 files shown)


app/abtests/Test.scala
```diff
@@ -2,26 +2,42 @@ package abtests
...
-    else Some(Allocation(this, variants.toList(id % variants.size)))
+  def canRun(req: Request[_]): Boolean
+
+  def allocate(id: Int, request: Request[_]): Option[Allocation] = {
+    if (idRange.contains(id) && canRun(request)) Some(Allocation(this, variants.toList(id % variants.size)))
+    else None
   }
 }
 
+case class SplitTest(name: String, audienceSize: Percentage, audienceOffset: Percentage, variants: Variant*) extends Test {
```
alexduf: It looks like you can keep the `Test` `case class` and define `canRun` as a function with a default value

alexduf (commented):
Looks good otherwise 👍

alexduf (approved):

PR #244, opened 2017-03-23
Keep amount parameter on redirect

alexduf:
keep the amount parameter in case the S&C team wants to use it

cc @guardian/contributions @Ap0c


dominickendrick:
Is this still needed?

alexduf:
I'm actually surprised it works without

dominickendrick (approved):

PR #247, opened 2017-04-04
Adding gulocal key and crt to allow nginx to work

dominickendrick:
This PR adds gulocal key and crt to the nginix setup to allow nginx to work.
This covers the scenario where you want to run nginx but have not got local `identity` setup


joelochlann (approved):
👍  🔑  🔒

PR #251, opened 2017-04-20
Support pay endpoint with JSON (for Apps)

alexduf:
In order to test the conversion of our app users using a native flow, we need an endpoint capable of ingesting json.
I'm transforming the /stripe/pay endpoint to be able to handle both a form and a json payload.

That really means the /stripe/pay endpoint will be used by both the apps and the frontend, and we need to think twice before drastically changing its behaviour (but adding more optional fields is fine!)

It's easier to read this PR commit per commit.

@guardian/contributions 

cc @maxspencer

### DIFF
--- a/app/controllers/StripeController.scala
+++ b/app/controllers/StripeController.scala
@@ -40,7 +40,7 @@ class StripeController(paymentServices: PaymentServices, stripeConfig: Config)(i
       Map(key -> value.identifier)
   }
 
-  case class SupportForm(
+  case class ContributionRequest(
     name: String,
     currency: Currency,
     amount: BigDecimal,
@@ -57,7 +57,7 @@ class StripeController(paymentServices: PaymentServices, stripeConfig: Config)(i
 
   )
 
-  val supportForm: Form[SupportForm] = Form(
+  val contributionForm: Form[ContributionRequest] = Form(
     mapping(
       "name" -> text,
       "currency" -> of[Currency],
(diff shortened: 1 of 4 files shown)


app/controllers/StripeController.scala
```diff
@@ -6,91 +6,31 @@ import java.time.Instant
...
-  val supportForm: Form[SupportForm] = Form(
-    mapping(
-      "name" -> text,
-      "currency" -> of[Currency],
-      "amount" -> bigDecimal(10, 2),
-      "email" -> email,
-      "token" -> nonEmptyText,
-      "marketing" -> boolean,
-      "postcode" -> optional(nonEmptyText),
-      "abTests" -> set(mapping(
```
dominickendrick, line 78: It there a reason to remove this abTest? Could it be made optional instead ?

Mullefa: @alexduf I think because these are inferred directly from the request?

Mullefa (approved):
LGTM (assuming it is OK to omit the `abTests` field)

app/controllers/StripeController.scala
alexduf: We parse the field but don't do anything with it. In other words it's dead code (unless I missed something)
It's now handled here https://github.com/guardian/contributions-frontend/blob/master/app/controllers/StripeController.scala#L113
I think @desbo did the work so he may be able to confirm

dominickendrick: Ok, that seems fine then 👍

dominickendrick (approved):

alexduf:
Thanks guys, I'll merge when tested end-to-end with andriod. (hopefully soon)

maxspencer:
🎉 nice one @alexduf

PR #256, opened 2017-04-25
Humanise tests v2

Mullefa:
__NB:__ _do not merge until the initial Humanise Test is complete_

cc @guardian/contributions 

This pull request removes the _Humanise Test_ and adds a new test - _Humanise Test V2_.

This new test has one non-control variant - `contributionCount` (screenshots below - the figure of 10,000 was hard coded for illustrative purposes).

__Desktop:__

contribution-count-desktop

__Mobile__:

contribution-count-mobile

PR #257, opened 2017-04-27
increase gap between contribution count and line below

Mullefa:
This pull request introduces a slight edit to the CSS in #256.

When the contribution count is smaller (than the hard coded 10,000) the vertical height between the comma in the number, and the text below is too small (as illustrated below).

css-edit

It now looks like this:

css-fix


joelochlann (approved):
Good spot 👍

PR #266, opened 2017-05-17
Add .nvmrc for version 7.8

joelochlann:
Some bits of our clientside build process require node version 7. This makes that fact explicit, so you can run `nvm use` in the contributions-frontend directory and it will set the right node version.

If you add the following to your `.bash_profile` then `nvm` will automatically do this every time you change directory (courtesy of https://gist.github.com/jusopi/fa6d03321991620778a2):

```
enter_directory(){
	if [ "$PWD" != "$PREV_PWD" ]; then
		PREV_PWD="$PWD";
		if [ -e ".nvmrc" ]; then
			nvm use;
		fi
	fi
}

export PROMPT_COMMAND="$PROMPT_COMMAND enter_directory;"
```

For zsh people there will be a different solution


joelochlann:
@guardian/contributions

desbo (approved):

PR #267, opened 2017-05-18
Add functionality to tag logs

Mullefa:
cc @guardian/contributions 

This pull request attempts to lay the groundwork for tagging log requests.

Log entries are currently tagged with `browser id` and `request id`.

When a request is made, an implicit `LoggingTags` instance is created and this is passed recursively to any functions which rely on this information for logging. 

Under-the-hood the [Mapped Diagnostic Context](https://logback.qos.ch/manual/mdc.html) which stores contextual data on a per-thread basis is updated and this is queried when the log call is made.

__Primary use case__: the [logback Sentry Appender](https://github.com/getsentry/raven-java/blob/master/raven-logback/src/main/java/com/getsentry/raven/logback/SentryAppender.java) queries the MDC to add tags before sending the data to Sentry. This will allow us to filter logs in Sentry by browser id or request id; or whatever tags we choose to add in the future.

__Secondary use case__ (not yet implemented): the log entries that are written to `logs/application.log` can be formatted to include this additional contextual information e.g. [Example 7.1](https://logback.qos.ch/manual/mdc.html)

### DIFF
--- a/app/monitoring/LoggingTags.scala
+++ b/app/monitoring/LoggingTags.scala
@@ -1,0 +1,56 @@
+package monitoring
+
+import java.util.UUID
+
+import com.typesafe.scalalogging.LazyLogging
+import org.slf4j.MDC
+import play.api.mvc.Request
+
+// Tags to be included in a log statement using Mapped Diagnostic Context (MDC).
+// A MDC can be used to supplement log messages with additional contextual information,
+// or provide tags to a Sentry Appender.
+//
+// References:
+// - https://logback.qos.ch/manual/mdc.html, and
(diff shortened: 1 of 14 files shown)


dominickendrick (approved):

app/monitoring/LoggingTags.scala
```diff
@@ -0,0 +1,56 @@
...
+}
+
+object LoggingTags {
+
+  val browserId = "browserId"
+  val requestId = "requestId"
+
+  val allTags = Seq(browserId, requestId)
+
+  // Means that if `implicit request =>` is used in an Action, an implicit SentryLoggingTags instance will be in scope.
```
rtyley: Missed a `SentryLoggingTags` rename there...

rtyley (approved):

PR #272, opened 2017-05-22
update the error messages related to the identity API

Mullefa:
cc @guardian/contributions 

One of the most common contribution front-end errors we get in Sentry is:

```
Unable to parse json from identity List((,List(ValidationError(List('use..
```

This pull request attempts to improve the logging around interaction with the Identity API.

### DIFF
--- a/app/services/IdentityService.scala
+++ b/app/services/IdentityService.scala
@@ -4,7 +4,9 @@ import com.typesafe.config.Config
 import models.{Autofill, IdentityId}
 import monitoring.LoggingTags
 import monitoring.TagAwareLogger
+import play.api.http.Status
 import play.api.libs.json._
+import play.api.libs.ws.WSResponse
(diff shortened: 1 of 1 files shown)


Mullefa:
@alexduf also just a quick check - the logic of the `name() / concatNames()` function will return `None` if either of the names are `None`. Is this what we want?

alexduf:
@Mullefa It is used to auto-fill the name on the payment screen.
When it comes to payment, either we have precise information and we can suggest the name, or we're not sure so it's probably better to abstain rather than using something that may be a pseudonym or an incorrect information.

We have loads of people with weird names that aren't related to their real names, and that's one way to prevent it. (check our DB, you'll find countless users named "ID251748" before that change was introduced)

That being said, I'm opened to any other solution

alexduf (approved):
After having a chat with you, all good 👍

PR #276, opened 2017-05-24
Hash browser

Mullefa:
cc @guardian/contributions, @adamnfish 

Currently we tag our log requests with certain values (e.g. browser id) to facilitate debugging errors in sentry. Motivated by GDPR classing browser ids as personal data (or something like that!), this pull request:

- includes a `hashValue` field in the `LoggingTag` trait, which is used to signify whether the respective tag value should be hashed when included in log entries
- provides a `Hasher` using the SHA-256 algorithm which is used for hashing logging tags when the respective `hashValue` equals `true`

### DIFF
--- a/app/controllers/PaypalController.scala
+++ b/app/controllers/PaypalController.scala
@@ -12,6 +12,7 @@ import com.paypal.api.payments.Payment
 import models._
 import monitoring.TagAwareLogger
 import monitoring.LoggingTags
+import monitoring.LoggingTagsProvider
 import play.api.libs.ws.WSClient
 import play.api.mvc._
 import services.PaymentServices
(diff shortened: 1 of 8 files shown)


davidfurey:
I don't know any of the detail of GDPR, but if GDPR does class a browser id as personal data, is a hash of it not also a unique identifier and therefore personal data?

Mullefa:
@adamnfish I recall you saying that storing a hash is ok, but to be honest I don't know exactly why.
Is it because the browser id could be linked back to the original reader, but the hash can't?

dominickendrick (approved):

adamnfish:
Thanks for the mention, everyone. I've spoken with a few of you IRL (and our data privacy expert). I'll try and summarise the conversation @Mullefa and I just had here:

In terms of our GDPR obligations, hashing the identifier like this changes this data from PII to "psedonymous". While this is an improvement, it doesn't actually meaningfully change the way the data needs to be treated. It'll still be subject to the same requirements (encryption, access control, auditing/monitoring etc) as PII. This means that we can't say "oh it's hashed" and stop there. This will still need the same care and attention as it would if we hadn't hashed it at all.

The preparation I emailed about last week is still the phase we're in now. To some extent there isn't heaps of point trying to put technical solutions in place just yet because we're still working out what shape those will take. That said, psedonymising this data is always a good idea so GDPR aside avoiding putting sensitive data in logs like these is great.

PR #283, opened 2017-06-05
Add Ticker to This Land Is Your Land page

sammorrisdesign-zz:
It looks like this...

image


dominickendrick:
@guardian/contributions

assets/stylesheets/contributions.scss
```diff
@@ -542,6 +542,91 @@ $currencies: gbp usd aud eur nzd cad;
     }
 }
 
+.ticker {
```
desbo: Could you move this to its own file and import it?

desbo:
The JS stuff here should really be implemented as a React component like the checkout, but I assume there's not enough time to do that before it goes out. We're a bit concerned about adding more non-ES6, non-React code to the project though.

We can merge it now, but at some point it should be rewritten - would you be up for doing that?

sammorrisdesign-zz:
@desbo Yeah ok I'll take a stab at that. Have never really used React before so would be a good excuse to learn. @dominickendrick had concerns about deploying on vote day, so I'll take a stab at the React version as we have a bit more time now.

desbo:
Nice! Hopefully there's a fair bit of material if you look at how the checkout works, but feel free to send over any questions

sammorrisdesign-zz:
So probably not my finest work. Not sure how much of a "right way" there is with React, but I've taken a stab at it, @desbo. It does work and I've tried to build it in a modular componenty way.

desbo:
Thanks a lot Sam, things are hectic here, we'll take a look on Monday.

assets/javascripts/src/modules/contribute.es6
```diff
@@ -27,12 +28,36 @@ export function init() {
...
+        });
+    }
+
     attachCurrencyListeners();
     attachErrorDialogListener();
     setOphanIds();
     autoFill();
 }
 
+function loadJSON(path, success) {
```
desbo: use `fetch`  (e.g. https://github.com/guardian/contributions-frontend/blob/8be7abddef50155d52503aab38ac98c95f7df6ca/assets/javascripts/src/modules/contribute.es6#L38)

assets/javascripts/src/components/Ticker.jsx
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    render() {
+        return (
+            <div className='ticker'>
+                <div className="ticker__wrapper l-constrained">
+                    <h3 className="ticker__count">${this.state.count.toLocaleString()}</h3>
+                    <h4 className="ticker__label">pledged so far</h4>
+                    <div className="ticker__progress">
+                        <div className="ticker__filled-progress" style={{transform: 'translateX(' + this.state.filledProgress + ')'}}></div>
```
dominickendrick: I would separate out the progress bar and the counter to be sub components. Then rather than having to re-render the entire html block every `30ms` we would only be updating the progress bar and counter elements, which would be more performant.

assets/javascripts/src/modules/contribute.es6
```diff
@@ -27,13 +28,27 @@ export function init() {
         container
     );
 
+    const tickerContainer = document.getElementById('ticker');
+    if (tickerContainer) {
+        fetch(tickerContainer.getAttribute('data-src'), {
```
dominickendrick, line 33: 👏

dominickendrick (approved):

PR #287, opened 2017-06-15
Reinstate testimonials survey

dominickendrick:
You thought it was back! But it's gone again! 💥  
We decided that its not going to be reinstated any time soon, so it was not worth putting the effort in to making a switch 🚫 🔘


joelochlann (approved):
Makes me think of Stephen's Last Night In Town

PR #291, opened 2017-06-21
Record ophan visitId

dominickendrick:
In preparation for sending acquisition events to Ophan, we need to have access to the Ophan VisitId, which is a ~~required~~ optional field. This PR passes this from the front end to the backend, and then through the Paypal and Stripe checkout flows. This would allow us to do more detailed session tracking prior to their contribution.

@guardian/contributions


alexduf:
@dominickendrick is it fine if apps aren't sending it yet? (as in, no crash?)

dominickendrick:
@alexduf yeah, I may remove this PR, as apps don't send it, and it's optional in ophan. I need to check how valuable it would actually be to record this.

alexduf:
We _can_ send it though!

desbo (approved):

PR #295, opened 2017-06-27
Record Stripe contributions to Ophan

dominickendrick:
This PR sends all contributions made via Stripe to Ophan by:
- Adding a new OphanService to send data to Ophan
- Refactoring Stripe controller to create metadata earlier, so that it can be passed to Ophan
- Modifying when abtest data is converted to `json` so we can more easily support Ophan data structure
- Rewiring the AppLoader to make Ophan service available via DI 

For future us, I'll explain why I needed to manually encode the query params that we send to Ophan here https://github.com/guardian/contributions-frontend/pull/295/files#diff-1c1dcac2764f790c720c451ff7d0c6fdR160

Currently `okhttp.HttpUrl` correctly encodes url parameters using the [RFC-3986](https://www.ietf.org/rfc/rfc3986.txt) spec which does __not__ require `{}` to be encoded.
But `Play` up to `2.5` uses the stricter (and now obsolete)  [RFC-2396](https://www.ietf.org/rfc/rfc2396.txt) which requires `{` and `}` to be percent encoded. This includes the version `2.4.3` which Ophan uses. This [issue has now been fixed](https://github.com/playframework/playframework/issues/6801) in the `Play` framework. But I'm not bumping the play version of Ophan as part of this change.

I've written some tests to check url encoding happens correctly.
I've also tested the payment processes locally

@guardian/contributions @davidfurey @rtyley @lindseydew @emma-p


app/services/OphanService.scala
```diff
@@ -0,0 +1,183 @@
...
+   paymentProvider: Option[PaymentProvider],
+   campaignCode: Option[Set[String]],
+   abTests: Set[Allocation],
+   countryCode: Option[String],
+   referrerPageViewId: Option[String],
+   referrerUrl: Option[String]
+) {
+  def toParams: Seq[(String, String)] = {
+    Seq(
+      "viewId" -> this.viewId,
```
davidfurey: You should be able to drop all of the `this.` prefixes

app/services/OphanService.scala
```diff
@@ -0,0 +1,183 @@
...
+      "browserId" -> this.browserId,
+      "product" -> this.product.stringValue,
+      "currency" -> this.currency,
+      "paymentFrequency" -> this.paymentFrequency.stringValue,
+      "amount" -> this.amount.toString
+    ) ++
+      this.visitId.map(visitId => "visitId" -> this.visitId.toString) ++
```
davidfurey: Since there are so mainly optional values, this might be clearer as:
```
List(
  "visitId" -> visitId.map(_.toString),
  "amountInGBP" -> amountInGBP.map(_.toString),
  ...
).collect { case (k, Some(v)) => k -> v }
```

app/services/StripeService.scala
```diff
@@ -83,6 +73,35 @@ class StripeService(
...
+    cmp: Option[String],
+    metadata: ContributionMetaData,
+    contributor: Contributor,
+    contributorRow: ContributorRow,
+    idUser: Option[IdentityId],
+    marketing: Boolean)
+    (implicit tags: LoggingTags): EitherT[Future, String, SavedContributionData] = {
+
+    // Fire and forget: we don't want to stop the user flow
+    idUser.map { id =>
```
davidfurey: use `.foreach` instead of `.map` if you are ignoring the result

app/services/OphanService.scala
dominickendrick: nice !

desbo (approved):

PR #297, opened 2017-07-04
Using slug, instead of name for AB tests in Ophan

dominickendrick:
This uses the `slug` instead of `name` when passing AB tests to Ophan

@guardian/contributions


jranks123 (approved):
w00t!

PR #298, opened 2017-07-04
Remove ELB session stickiness

desbo:
This is preventing caching in Fastly (due to the Set-Cookie header being
set by the ELB). We need caching for the moment because we're about to
send a large amount of traffic to the app from the PayPal-Epic iframe.

@guardian/contributions

### DIFF
--- a/cloud-formation/cfn.yaml
+++ b/cloud-formation/cfn.yaml
@@ -176,9 +176,6 @@
         ConnectionDrainingPolicy:
           Enabled: true
           Timeout: 60
-        LBCookieStickinessPolicy:
-          - CookieExpirationPeriod: 1800
-            PolicyName: CookieBasedPolicy
     LoadBalancerSecurityGroup:
       Type: AWS::EC2::SecurityGroup
       Properties:


philwills (approved):
Sticky is icky

alexduf:
The reason I had to enable it was because the application could receive requests from the two versions of the app at the same time during deployment.
That usually was translated into client side errors right after or during the deployment.

It's even more important as the application is processing more traffic than it use to.

I think that could be avoided by using asset versioning.

With that in mind, I do dislike session stickiness as well.

desbo:
@alexduf as I understand it we are versioning assets: https://github.com/guardian/contributions-frontend/blob/master/conf/routes#L10 - have I missed something?

alexduf:
That's true it fixes part of the issue but there's still that case that can cause trouble:

1. Deployment is ongoing, two versions v1 and v2 are running in parallel
2. The client receives index.html from v2
3. The client requests the assets linked to the v2 page, but the ELB sends the request to a v1 server
4. v1 server can't find a v2 asset and responds with an HTTP 404
5. the page is broken

Granted: it's not always happening, and I hate to slow down work, but I believe this PR may (sporadically) break the contribution page. 

I think frontend came up with a solution that can solve the issue: hosting the assets on s3, but that means more work for you. (I think) that could be achieved by:
 - modifying the build (sbt or webpack) to package a new riff-raff package
 - modifying riff-raff.yaml to upload that package to s3
 - modifying the fastly configuration to route to s3 when accessing the assets path.
 
I'm happy to have a chat if you want, and I'm happy if you want to merge anyway, but I think it's worth considering the above solution before merging.

dominickendrick:
Fastly does support a method for retrying on 404's https://docs.fastly.com/guides/performance-tuning/checking-multiple-backends-for-a-single-request but it would be better to fix the problem by deploying and reading assets from s3

desbo:
For reference, we're going with the S3 approach, as seen (in part) in #299

PR #299, opened 2017-07-05
Deploy assets to S3

desbo:
This change adds: 

* a build script to upload assets to the riff-raff bucket
* a new deployment step (stage?) to riff-raff.yml which moves the assets to an S3 bucket accessible by the app instances

When this is released, we'll need to update Fastly to redirect any asset requests to the S3 bucket. 
With all of that in place, assets will be served from S3 and we can remove ELB stickiness _and_ enable caching on our templates without having to worry about 404s on static resource requests. 

I've also removed a build step to clear the NPM cache, which should significantly improve build times.

@guardian/contributions


grunt-tc
```diff
@@ -9,10 +9,9 @@ set -o errexit
 
 ################################################################################
 # Build the static asset files for the project
-# Files will now be found in Public folder
+# Files will now be found in the public folder
 ################################################################################
 
-npm cache clean
```
dominickendrick, line 15: Is this not needed ?

desbo: I don't see why it should be, it's just a disk cache of the package at a specific version. Re-downloading every NPM package on every build adds a good 3-4 minutes each time.

From https://docs.npmjs.com/cli/cache: 
> All data that passes through the cache is fully verified for integrity on both insertion and extraction. Cache corruption will either trigger an error, or signal [...] that the data must be refetched, which it will do automatically. For this reason, it should never be necessary to clear the cache for any reason other than reclaiming disk space

dominickendrick (approved):

PR #302, opened 2017-07-07
Add logging and cloudwatch

michaelwmcnamara:
This change adds logging and cloudwatch metrics to allow for production monitoring of payment systems and improved logging for investigation of issues.

Confirmed payments in Stripe and PAypal(sandbox) work ok when run on local

@guardian/contributions


app/controllers/Contributions.scala
```diff
@@ -106,6 +108,8 @@ class Contributions(paymentServices: PaymentServices, addToken: CSRFAddToken) ex
       .map(mobileRedirectUrl)
       .filter(_ => request.isIos)
 
+    info(s"thank you page displayed for paypal payment id: ${request.id}")
+    putThankYouPage()
```
dominickendrick: Can this and all other cloudwatch metrics calls be called `logThankYouPage`. I was a little confused about what this did from the name. Also the brackets are not required.

app/controllers/StripeController.scala
```diff
@@ -121,17 +125,23 @@ class StripeController(paymentServices: PaymentServices, stripeConfig: Config, o
     }
 
     createCharge.map { charge =>
+      val platform = request.platform.getOrElse("unknown")
       val metadata = createMetaData(charge)
       storeMetaData(metadata) // fire and forget. If it fails we don't want to stop the user
       recordToOphan(metadata) // again, fire and forget.
-      Ok(Json.obj("redirect" -> thankYouUri))
+      Ok({info(s"Stripe payment successful for request id: ${request.id.toString}, \n\t from platform: $platform")
```
dominickendrick: Can we split this across multiple lines so it reads more clearly

app/controllers/StripeController.scala
```diff
@@ -121,17 +125,23 @@ class StripeController(paymentServices: PaymentServices, stripeConfig: Config, o
...
+        putStripePaymentSuccess(platform)
+        Json.obj("redirect" -> thankYouUri)})
         .addingToSession("charge_id" -> charge.id)
         .addingToSession("amount" -> contributionAmount.show)
         .setCookie[ContribTimestampCookieAttributes](Instant.ofEpochSecond(charge.created).toString)
     }.recover {
-      case e: Stripe.Error => BadRequest(Json.toJson(e))
+      case e: Stripe.Error => {warn(s"Payment failed for request id ${request.id.toString}, \n\t from platform: $platform, \n\t with code: ${e.code}, \n\t decline code: ${e.decline_code}, \n\t and error message: ${e.message}.")
```
desbo: Could you insert a new line after the `{`?

desbo:
Looks fine to me apart from some formatting issues, would you be able to run "Reformat code" in intellij on the changed/new files?

app/controllers/Contributions.scala
Mullefa: I know it's subjective, but I think the brackets are helpful to indicate that evaluating the expression has a side-effect.

app/monitoring/CloudWatch.scala
```diff
@@ -0,0 +1,65 @@
...
+import com.gu.aws.CredentialsProvider
+import monitoring.CloudWatch.cloudwatch
+
+trait CloudWatch extends TagAwareLogger {
+  val stage: String
+  val application : String
+  lazy val stageDimension: Dimension = new Dimension().withName("Stage").withValue(stage)
+  def mandatoryDimensions:Seq[Dimension] = Seq(stageDimension)
+
+  trait LoggingAsyncHandler extends AsyncHandler[PutMetricDataRequest, PutMetricDataResult]
```
Mullefa: could we keep the styling consistent with the rest of the code base?  e.g.1. `trait Foo {` rather than having `{` on a new line; e.g.2 being explicit on the return type of functions that return a unit (`def foobar(): Unit = ...`)

Mullefa: This could just be a object which extends `AsyncHandler` and `LazyLogging` and is defined outside of the `CloudWatch` trait.

app/monitoring/CloudWatch.scala
```diff
@@ -0,0 +1,65 @@
...
+  object LoggingAsyncHandler extends LoggingAsyncHandler
+
+
+  def put(name : String, count: Double, extraDimensions: Dimension*): Future[PutMetricDataResult] = {
+    val metric =
+      new MetricDatum()
+        .withValue(count)
+        .withMetricName(name)
+        .withUnit("Count")
+        .withDimensions((mandatoryDimensions ++ extraDimensions): _*)
```
Mullefa: inner brackets not needed

app/monitoring/CloudWatch.scala
```diff
@@ -0,0 +1,65 @@
...
+
+    val request = new PutMetricDataRequest().
+      withNamespace(application).withMetricData(metric)
+
+    cloudwatch.putMetricDataAsync(request, LoggingAsyncHandler)
+  }
+
+  def put(name: String, count: Double, responseMethod: String) {
```
Mullefa: the method never seems to be used - is it needed?

app/monitoring/CloudWatch.scala
```diff
@@ -0,0 +1,65 @@
...
+    {
+      logger.trace("CloudWatch PutMetricDataRequest - success")
+      CloudWatchHealth.hasPushedMetricSuccessfully = true
+    }
+  }
+
+  object LoggingAsyncHandler extends LoggingAsyncHandler
+
+
+  def put(name : String, count: Double, extraDimensions: Dimension*): Future[PutMetricDataResult] = {
```
Mullefa: I don't think the following `put()` method (line 48) is needed, in which case the `extraDimensions` argument can be omitted.

app/monitoring/CloudWatch.scala
```diff
@@ -0,0 +1,65 @@
...
+import com.amazonaws.regions.Regions.EU_WEST_1
+import com.amazonaws.services.cloudwatch.AmazonCloudWatchAsyncClient
+import com.amazonaws.services.cloudwatch.model.{Dimension, MetricDatum, PutMetricDataRequest, PutMetricDataResult}
+import com.gu.aws.CredentialsProvider
+import monitoring.CloudWatch.cloudwatch
+
+trait CloudWatch extends TagAwareLogger {
+  val stage: String
+  val application : String
+  lazy val stageDimension: Dimension = new Dimension().withName("Stage").withValue(stage)
```
Mullefa: if you used an abstract class with parameters or [delayed trait initialisation](https://stackoverflow.com/questions/16251379/delaying-trait-initialization) with respect to `stage` and `application` you could omit the `lazy val`, in which case `mandatoryDimensions` could be a regular `val` as well.

app/monitoring/CloudWatch.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  lazy val cloudwatch = AmazonCloudWatchAsyncClient.asyncBuilder
+    .withCredentials(CredentialsProvider)
+    .withRegion(EU_WEST_1).build()
+
+}
+
+
+object CloudWatchHealth {
+  var hasPushedMetricSuccessfully = false
```
Mullefa: this value never seems to be read - is it needed?

app/monitoring/CloudWatch.scala
Mullefa: Could the full class path for`Future` be used or it be imported as `JFuture`? otherwise I think someone will just assume it's a Scala future

app/monitoring/ContributionMetrics.scala
```diff
@@ -0,0 +1,94 @@
...
+
+import configuration.Config
+import play.api.libs.json.Json
+
+trait ContributionMetrics extends CloudWatch{
+
+  val stage: String = Config.stage
+  val application = "contributions" // This sets the namespace for Custom Metrics in AWS (see CloudWatch)
+
+  def putStripePaymentAttempt(platform: String): Unit ={
```
Mullefa: Is it ok we are recording platform for Stripe, but not for Paypal?

app/controllers/Contributions.scala
michaelwmcnamara: Hi Dom, 
Thanks for this! The cloudwatch metric functions have been renamed "log" instead of put - as per request.

app/controllers/StripeController.scala
michaelwmcnamara: Done, hope this format is ok.

app/controllers/StripeController.scala
michaelwmcnamara: Thanks - good catch!

app/monitoring/CloudWatch.scala
michaelwmcnamara: Good point! 
Thanks :-)

app/controllers/PaypalController.scala
```diff
@@ -88,87 +98,10 @@ class PaypalController(ws: WSClient, paymentServices: PaymentServices, checkToke
...
-      case JsString(id) => CountryGroup.byId(id).map(JsSuccess(_)).getOrElse(JsError("invalid CountryGroup id"))
-      case _ => JsError("invalid value for country group")
-    }
-  }
-
-  private def capAmount(amount: BigDecimal, currency: Currency): BigDecimal = amount min MaxAmount.forCurrency(currency)
-
   def authorize = checkToken {
     NoCacheAction.async(parse.json[AuthRequest]) { implicit request =>
+      info(s"Attempting to obtain paypal auth url.")
```
Mullefa: _attempting to create Paypal payment_ may be clearer - the payment is then used to derive the approval url and payment id

app/controllers/PaypalController.scala
```diff
@@ -188,16 +121,24 @@ class PaypalController(ws: WSClient, paymentServices: PaymentServices, checkToke
       payment.subflatMap(AuthResponse.fromPayment).fold(
         err => {
           error(s"Error getting PayPal auth url: $err")
+          logPaypalAuthFailure()
           InternalServerError("Error getting PayPal auth url")
```
Mullefa: _auth response_ is more accurate

app/controllers/PaypalController.scala
```diff
@@ -188,16 +121,24 @@ class PaypalController(ws: WSClient, paymentServices: PaymentServices, checkToke
...
         },
-        authResponse => Ok(Json.toJson(authResponse))
+        authResponse => {
+          info(s"Paypal payment auth url successfully obtained.")
```
Mullefa: _auth response_ is more accurate

app/controllers/PaypalController.scala
```diff
@@ -223,13 +167,19 @@ class PaypalController(ws: WSClient, paymentServices: PaymentServices, checkToke
...
-      "marketingOptIn"->boolean
-    )(MetadataUpdate.apply)(MetadataUpdate.unapply)
-  )
+  implicit val UriWrites = new Writes[Uri] {
+    override def writes(uri: Uri): JsValue = JsString(uri.toString)
+  }
+
+  implicit val AuthResponseWrites = Json.writes[AuthResponse]
+
+  implicit val CountryGroupReads = new Reads[CountryGroup] {
```
Mullefa: this should be defined in `views.support.CountryGroupImplicits`

app/controllers/PaypalController.scala
```diff
@@ -223,13 +167,19 @@ class PaypalController(ws: WSClient, paymentServices: PaymentServices, checkToke
...
-  val metadataUpdateForm: Form[MetadataUpdate] = Form(
-    mapping(
-      "marketingOptIn"->boolean
-    )(MetadataUpdate.apply)(MetadataUpdate.unapply)
-  )
+  implicit val UriWrites = new Writes[Uri] {
+    override def writes(uri: Uri): JsValue = JsString(uri.toString)
+  }
+
+  implicit val AuthResponseWrites = Json.writes[AuthResponse]
```
Mullefa: this should be defined in the `AuthResponse` companion object. Also, being a value, it should be named `authResponseWrites`

Mullefa: ignore this - I thought it was introduced in the pull request,  but actually it wasn't

app/controllers/PaypalController.scala
Mullefa: ignore this - I thought it was introduced in the pull request,  but actually it wasn't

michaelwmcnamara:
Closing this pull request.
The reason being - we want to minimise risk in code changes until we have proper logging and monitoring in place. This PR includes significant refactoring as well as the addition of cloudwatch metrics and improved logging.
It will be split into multiple pr's.
The cloudwatch metrics and logging will now be added in https://github.com/guardian/contributions-frontend/pull/303

PR #303, opened 2017-07-13
Add cloudwatch metrics and logging - expurgated version

michaelwmcnamara:
This pull requests adds improves logging and adds cloudwatch metrics for Stripe and Paypal payments.

Testing:
Confirmed that payments work locally

@guardian/contributions

### DIFF
--- a/app/controllers/StripeController.scala
+++ b/app/controllers/StripeController.scala
@@ -16,9 +16,8 @@ import controllers.forms.ContributionRequest
 import cookies.ContribTimestampCookieAttributes
 import cookies.syntax._
 import models._
-import monitoring.LoggingTagsProvider
+import monitoring.{CloudWatchMetrics, LoggingTagsProvider, TagAwareLogger}
 import org.joda.time.DateTime
-import monitoring.TagAwareLogger
 import play.api.libs.json._
 import play.api.mvc._
 import services.Ophan.OphanResponse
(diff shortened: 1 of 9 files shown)


dominickendrick (approved):
Looks great, a lot easier to review : ) Lets discuss our testing strategy before we release this

app/controllers/StripeController.scala
```diff
@@ -129,7 +133,11 @@ class StripeController(paymentServices: PaymentServices, stripeConfig: Config, o
         .addingToSession("amount" -> contributionAmount.show)
         .setCookie[ContribTimestampCookieAttributes](Instant.ofEpochSecond(charge.created).toString)
     }.recover {
-      case e: Stripe.Error => BadRequest(Json.toJson(e))
+      case e: Stripe.Error => {
```
Mullefa, line 136: I'm wondering if we want to recover from any non-fatal error here? In the context of this pull request it'd mean we'd log any Stripe payment failure with Cloud Watch. More generally, we'd send a specific error to the client instead of an Internal Server Error (which I don't think is currently handled).

michaelwmcnamara: Thats a good idea, but I think we should do it in a different PR. Remember the aim of this PR is to get the logging and metrics out without changing any functionality, so that we minimise the risk of deploying it.

app/monitoring/CloudWatchMetrics.scala
```diff
@@ -0,0 +1,125 @@
...
+import com.amazonaws.services.cloudwatch.AmazonCloudWatchAsync
+import com.amazonaws.services.cloudwatch.model.{Dimension, MetricDatum, PutMetricDataRequest}
+import configuration.Config
+import play.api.libs.json.Json
+
+
+/**
+  * Created by mmcnamara on 12/07/2017.
+  */
+class CloudWatchMetrics(cloudWatchClient: AmazonCloudWatchAsync) extends TagAwareLogger {
```
Mullefa, line 12: No longer needs to extend `TagAwareLogger`.

app/monitoring/CloudWatchMetrics.scala
```diff
@@ -0,0 +1,125 @@
...
+
+  val stage: String = Config.stage
```
Mullefa, line 14: Minor point, but since we are listing the dependency on the cloud watch client explicitly, we might as well as well do the same for `stage` and provide it as a class parameter.

PR #305, opened 2017-07-18
adding some metrics to handle payments from apps 

michaelwmcnamara:
1) Payments from apps use the same payment functions as payments for stripe from web pages, but are redirected to a different place for payment successes.
As a result, I needed to add extra metrics to track payments from apps so that I can avoid giving false alarms.

2) Thank you page is reached from both paypal and stripe payments. Because the navigation from paypal is a manual click from the post-payment page, it is reasonable to assume a large number of users will not bother to click through to it.
This is a problem for tracking stripe payments as it is the page shown automatically for stripe users that  have made a successful payment. The fact that a paypal user may or may not click through to this page introduces a chaotic element into the count of thank-you page displays.
As a result, I have added a string detailing the payment method used to the request.session object that is accessible from all pages. 
This enables me to log which payment method a user used when reaching the thank you page, meaning I can better target the relation between stripe payment attempts and thank you page views reached via a stripe payment.


app/controllers/PaypalController.scala
```diff
@@ -80,7 +80,8 @@ class PaypalController(ws: WSClient, paymentServices: PaymentServices, checkToke
         case Accepts.Html() =>
           val amount = paypalService.paymentAmount(payment)
           val email = payment.getPayer.getPayerInfo.getEmail
-          val session = List("email" -> email) ++ amount.map("amount" -> _.show)
+          val paymentMethod = "paypal"
```
dominickendrick: This could just be hard coded into the list below.

dominickendrick (approved):

Mullefa:
It [seems](http://blog.librato.com/posts/tagged-metrics) that CloudWatch dimensions are mapped nicely to Librato tags. In which case, I think it'd be better to model `payment method` and `platform` as dimensions instead of combining them with previous metrics to create new ones.

app/actions/CommonActions.scala
```diff
@@ -52,9 +53,9 @@ object CommonActions {
   }
 
   implicit class MobileSupportRequest[A](val request: Request[A]) extends AnyVal {
-    def platform: Option[String] = request.getQueryString("platform") orElse request.session.get("platform")
-    def isAndroid: Boolean = platform.contains("android")
-    def isIos: Boolean = platform.contains("ios")
+    def platform: String = request.getQueryString("platform").orElse(request.session.get("platform")).getOrElse("web")
```
Mullefa: `platform` is also derived from a request [here](https://github.com/guardian/contributions-frontend/pull/305/files#diff-0ad4a3f0bc037938eadbb8872d92a4b3R65). @alexduf could this method be used instead?

app/controllers/StripeController.scala
```diff
@@ -32,9 +33,8 @@ class StripeController(paymentServices: PaymentServices, stripeConfig: Config, o
   // THIS ENDPOINT IS USED BY BOTH THE FRONTEND AND THE MOBILE-APP
   def pay = (NoCacheAction andThen MobileSupportAction andThen ABTestAction)
     .async(BodyParsers.jsonOrMultipart(ContributionRequest.contributionForm)) { implicit request =>
-    val platform = request.platform.getOrElse("unknown")
-    info(s"A Stripe payment is being attempted with request id: ${request.id}. \n\t Request is from platform: $platform.")
-    cloudWatchMetrics.logStripePaymentAttempt(platform)
+    info(s"A Stripe payment is being attempted with request id: ${request.id}, from platform: ${request.platform}.")
+    cloudWatchMetrics.logPaymentAttempt(PaymentProvider.Stripe, request.platform)
```
Mullefa, line 37: @michaelwmcnamara just realised, in this controller the `form` takes precedent when deciding what the `platform` is. Therefore the value of platform should be `form.platform.getOrElse(request.platform)`

PR #307, opened 2017-07-21
Allow cross origin requests from support.theguardian.com

davidfurey:
This will allow us to run a test on support.theguardian.com offering monthly recurring and one-off contributions, without having to interrupt the users flow to send them to contribute.theguardian.com.


desbo (approved):

app/controllers/StripeController.scala
```diff
@@ -29,6 +29,10 @@ import scala.concurrent.{ExecutionContext, Future}
 class StripeController(paymentServices: PaymentServices, stripeConfig: Config, ophanService: OphanService, cloudWatchMetrics: CloudWatchMetrics)(implicit ec: ExecutionContext)
   extends Controller with Redirect with TagAwareLogger with LoggingTagsProvider {
 
+  def payOptions = CachedAction { request =>
+    NoContent.withHeaders(("Vary" -> "Origin") :: corsHeaders(request): _*)
```
desbo, line 34: I think you need to include `Access-Control-Allow-Methods` here?

davidfurey: GET, HEAD, and POST are allowed by default, so I don't need to specify

PR #308, opened 2017-07-21
Adding complete event to ab test data

dominickendrick:
This adds the `complete` part to the ab test data we pass as part of the acquisition event. This is required by Ophan's ab test parsing model!

@guardian/contributions

### DIFF
--- a/app/abtests/Test.scala
+++ b/app/abtests/Test.scala
@@ -67,7 +67,7 @@ case class Variant(name: String) {
 
 case class Allocation(test: Test, variant: Variant) {
   val toOphanJson: JsObject = {
-    Json.obj(test.slug -> Json.obj("variantName" -> variant.slug))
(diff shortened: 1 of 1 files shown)


desbo:
Would it be worth implementing this further downstream? In the context of `Allocation` a complete event doesn't really make sense, so it might be worth handling the field not existing (and defaulting to `false`) somewhere in Ophan. It would also mean other producers of this type of event wouldn't have to create a dummy value.

dominickendrick:
@desbo yeah I agree, When parsing AB test data, currently Ophan requires you to set a value, rather than defaulting it to false. I could also parse this data differently in Ophan ...

dominickendrick:
I suggest that we merge this change, and then fix the issue upstream in ophan in a later PR. This is purely to make the release of current changes slightly easier.

desbo (approved):

PR #311, opened 2017-07-25
Unit tests added for the Paypal controller

Mullefa:
cc @guardian/contributions 

Unit tests for the Paypal controller. I've had to refactor the code in this controller so the methods can be exposed for unit tests.

_Have you gone through the contributions flow for all test variants ?_ __Yes, on web__

### DIFF
--- a/app/controllers/PaypalController.scala
+++ b/app/controllers/PaypalController.scala
@@ -26,6 +26,7 @@ import scala.util.Try
 
 class PaypalController(paymentServices: PaymentServices, checkToken: CSRFCheck, cloudWatchMetrics: CloudWatchMetrics)(implicit ec: ExecutionContext)
   extends Controller with Redirect with TagAwareLogger with LoggingTagsProvider {
+  import PaypalController._
 
   def executePayment(
     countryGroup: CountryGroup,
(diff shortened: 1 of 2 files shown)


app/controllers/PaypalController.scala
```diff
@@ -248,30 +141,191 @@ class PaypalController(paymentServices: PaymentServices, checkToken: CSRFCheck,
...
+      implicit val authResponseWrites: Writes[AuthResponse] = Json.writes[AuthResponse]
+    }
+  }
+
+  // Class containing methods used to implement the execute payment endpoint.
+  // Factored out here to facilitate unit testing.
+  class ExecutePaymentUtils(paypalService: PaypalService, countryGroup: CountryGroup, cloudWatchMetrics: CloudWatchMetrics)
+    extends PaypalControllerUtils {
+
+    def executePayment(paymentId: String, payerId: String)(implicit request: ABTestRequest[AnyContent]): EitherT[Future, String, Payment] =
```
joelochlann, line 267: As discussed - you could probably get rid of this one and just call `paypalService.executePayment` directly

joelochlann (approved):
Nice work Guy - all looks good from what I can tell! And moving those functions out of the controllers has cleaned up the code.

I might get a second opinion on the usefulness of testing bit of logic which are so tied to a specific controller, when we have an end-to-end test for that controller. Seems like it could be valuable as an extra safety net but I do find it quite hard to imagine a problem which would be caught by one of these tests but not by the end-to-end test. But someone with more experience than me may well be able to!

desbo:
@joelochlann I think it's worth having these tests alongside the end-to-end ones as they cover cases that the Selenium ones won't, e.g. payment capping and errors. The selenium ones only cover the most fundamental path of completing a payment through card and PayPal, so anything extra is a bonus IMO.

@Mullefa you could probably look to merge this now that the end-to-end tests are running – I suspect the conflicts might not be much fun though 8(

Mullefa:
yeah, gonna wait for @alexduf to merge his most recent pull request, then will resolve the conflicts 🙈

dominickendrick:
Can this be merged ?

Mullefa:
yes, but conflicts are going to be a real pain to resolve. I probably won't have time to do it this week.

PR #315, opened 2017-08-01
A better id field in logs for debugging

michaelwmcnamara:
This updates the "request id" field in logs to instead use a fragment of the csrf-token cookie.
This is because the request id changes during a payment for a user making it useless for debugging.
This cookie fragment is consistent through the entire payment flow for a user, so its easier to tell where something has gone wrong when debugging.

yes

### DIFF
--- a/app/actions/CommonActions.scala
+++ b/app/actions/CommonActions.scala
@@ -56,14 +56,14 @@ object CommonActions {
   implicit class MobileSupportRequest[A](val request: Request[A]) extends AnyVal {
 
     private def _platform = request.getQueryString("platform").orElse(request.session.get("platform")).getOrElse("web")
-    private def _sessionId = request.cookies.get("PLAY_SESSION").map(_.value).getOrElse("unknown")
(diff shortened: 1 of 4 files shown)


app/actions/CommonActions.scala
```diff
@@ -56,12 +56,15 @@ object CommonActions {
   implicit class MobileSupportRequest[A](val request: Request[A]) extends AnyVal {
 
     private def _platform = request.getQueryString("platform").orElse(request.session.get("platform")).getOrElse("web")
+    private def _sessionId = request.cookies.get("gu.contributions.csrf-token").map(_.value).getOrElse("Unknown")
```
desbo, line 59: It would be good not to couple this to the CSRF cookie (which has an uncertain future). I think the Play session would be the place to do it – storing a UUID there should do the trick. 

https://www.playframework.com/documentation/2.5.x/ScalaSessionFlash#Storing-data-in-the-Session

michaelwmcnamara:
Will attempt this again using the suggested Play session when I return from leave

PR #318, opened 2017-08-07
Paypal mobile preparation

alexduf:
I was about to create a PR for the paypal mobile support but it was about 400 lines of code.

400 being to big to be digestible, I split it in two:
 - refactor of the existing code base (the present PR)
 - support for mobile payments via paypal

This PR focuses on:
 - extracting models from the paypal controller file in order to remove noise, and store them separately
 - refactor the test to ensure we test that the metadata is stored

cc @guardian/contributions cc @mohammad-haque

### DIFF
--- a/app/controllers/forms/AuthRequest.scala
+++ b/app/controllers/forms/AuthRequest.scala
@@ -1,0 +1,81 @@
+package controllers.forms
+
+import com.gu.i18n.CountryGroup
+import com.netaporter.uri.Uri
+import com.paypal.api.payments.Payment
+import play.api.libs.json.Reads.min
+import play.api.libs.json._
+import play.api.libs.functional.syntax._
+import utils.JsonUtils._
+
+import scala.util.Try
+
+
+case class AuthRequest private (
+  countryGroup: CountryGroup,
+  amount: BigDecimal,
+  cmp: Option[String],
+  intCmp: Option[String],
+  refererPageviewId: Option[String],
+  refererUrl: Option[String],
+  ophanPageviewId: Option[String],
+  ophanBrowserId: Option[String],
+  ophanVisitId: Option[String]
+)
+
+object AuthRequest {
+  /**
+    * We need to ensure there's no fragment in the URL here, as PayPal appends some query parameters to the end of it,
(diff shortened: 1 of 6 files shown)


alexduf:
a second PR will follow when this one is merged, based on the branch `paypal-mobile2`

conf/application.conf
```diff
@@ -1,5 +1,5 @@
 play.application.loader = "wiring.AppLoader"
-
+play.http.secret.key="a-very-secret-dev-key"
```
desbo, line 2: did you mean to include this?

app/controllers/forms/AuthRequest.scala
```diff
@@ -0,0 +1,81 @@
+package controllers.forms
```
joelochlann: Nice to see this moved out of controllers - is `forms` the right place for it though? Since it's not a form?

test/controllers/PaypalControllerSpec.scala
```diff
@@ -74,23 +112,31 @@ class PaypalControllerSpec extends PlaySpec
...
+          .thenReturn(EitherT.pure[Future, String, Payment](mockPaypalPayment))
+      }
 
-      val result: Future[Result] = executePayment(fakeRequest)
+      val result: Future[Result] = executePayment(fixture.controller)(fakeRequest)
 
       status(result).mustBe(303)
       redirectLocation(result).mustBe(Some("/uk/post-payment"))
+
+      fixture.storeMetaDataCalled(1)
```
joelochlann: It was quite non-obvious to me at first that this is another condition for the test to pass. It would be much clearer to me if it read something like

```scala
wasMetaDataStored().mustBe(true)
```

Is it easy to refactor it so it reads more like that?

conf/application.conf
alexduf: As we just chatted about:
Yes but only to simplify the set-up for anyone new as that property is necessary to start play. It's not really related to that PR but I figured it shouldn't hurt

joelochlann (approved):
Looks good!

alexduf:
tested on code ✔️

PR #319, opened 2017-08-08
Add username to Stripe form

desbo:
Hopefully the final thing to get the tests running against PROD.


joelochlann (approved):
Looks good

desbo:
Just to note, the fail status I just removed was only because the test wasn't properly at this point.

PR #325, opened 2017-08-09
Integrate with paypal on mobile

alexduf:
When using the paypal SDK on android, the payment cannot be executed, but instead need to be "captured"

This PR provides an endpoint to capture a payment. It also provided a more sane type for handling paypal errors to enable us to send a reasonable error back to the client side if necessary.

Finally, it also adds support for the capture events at the webhook level.

Before merging this PR, I will therefore need to modify the webhook configuration in the paypal console to enable the capture events.

@guardian/contributions 
cc @mohammad-haque

### DIFF
--- a/app/controllers/PaypalController.scala
+++ b/app/controllers/PaypalController.scala
@@ -47,7 +47,7 @@ class PaypalController(paymentServices: PaymentServices, corsConfig: CorsConfig,
     val paypalService: PaypalService = paymentServices.paypalServiceFor(request)
 
     paypalService.capturePayment(captureBody.paymentId)
-      .flatMap { payment =>
+      .map { payment =>
         paypalService.storeMetaData(
           paymentId = payment.getId,
           testAllocations = request.testAllocations,
@@ -60,7 +60,10 @@ class PaypalController(paymentServices: PaymentServices, corsConfig: CorsConfig,
           idUser = captureBody.idUser,
           platform = Some(captureBody.platform),
           ophanVisitId = None
-        )
+        ).leftMap { error =>
+          logger.error(s"Unable to store the metadata while capturing the payment. Continuing anyway. Error: $error")
+        }
+        payment
       }
       .fold(error => {
         logger.error(s"Unable to capture the payment: $error")
--- a/app/controllers/PaypalController.scala
+++ b/app/controllers/PaypalController.scala
@@ -8,11 +8,11 @@ import cookies.syntax._
 import com.gu.i18n.{CountryGroup, Currency}
 import com.paypal.api.payments.Payment
 import configuration.{CorsConfig, SupportConfig}
-import controllers.httpmodels.{AuthRequest, AuthResponse}
+import controllers.httpmodels.{AuthRequest, AuthResponse, CaptureRequest}
 import models._
 import monitoring._
 import play.api.mvc._
-import services.{PaymentServices, PaypalApiConfig}
+import services.{PaymentServices, PaypalService}
 import play.api.data.Form
 import utils.MaxAmount
 import play.api.libs.json._
@@ -40,6 +40,29 @@ class PaypalController(paymentServices: PaymentServices, corsConfig: CorsConfig,
         "Access-Control-Allow-Credentials" -> "true"
       )
     }
+  }
+
+  def capturePayment = (NoCacheAction andThen ABTestAction).async(parse.json[CaptureRequest]) { implicit request =>
+    val captureBody = request.body
+    val paypalService: PaypalService = paymentServices.paypalServiceFor(request)
+
+    paypalService.capturePayment(captureBody.paymentId)
+      .flatMap { payment =>
+        paypalService.storeMetaData(
+          paymentId = payment.getId,
+          testAllocations = request.testAllocations,
+          cmp = captureBody.cmp,
+          intCmp = captureBody.intCmp,
+          refererPageviewId = captureBody.refererPageviewId,
+          refererUrl = captureBody.refererUrl,
+          ophanPageviewId = captureBody.ophanPageviewId,
+          ophanBrowserId = captureBody.ophanBrowserId,
+          idUser = captureBody.idUser,
+          platform = Some(captureBody.platform),
+          ophanVisitId = None
+        )
+      }
+      .fold({_ => InternalServerError}, {_ => Ok})
   }
 
   def executePayment(
--- a/app/controllers/PaypalController.scala
+++ b/app/controllers/PaypalController.scala
@@ -67,7 +67,7 @@ class PaypalController(paymentServices: PaymentServices, corsConfig: CorsConfig,
       }
       .fold(error => {
         logger.error(s"Unable to capture the payment: $error")
-        InternalServerError(error.message)
+        InternalServerError(Json.toJson(error))
       }, {_ => Ok})
   }
 
--- a/app/controllers/PaypalController.scala
+++ b/app/controllers/PaypalController.scala
@@ -60,14 +60,14 @@ class PaypalController(paymentServices: PaymentServices, corsConfig: CorsConfig,
           idUser = captureBody.idUser,
           platform = Some(captureBody.platform),
           ophanVisitId = None
-        ).leftMap { error =>
-          logger.error(s"Unable to store the metadata while capturing the payment. Continuing anyway. Error: $error")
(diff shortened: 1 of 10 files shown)


app/controllers/PaypalController.scala
```diff
@@ -42,6 +42,35 @@ class PaypalController(paymentServices: PaymentServices, corsConfig: CorsConfig,
...
+          idUser = captureBody.idUser,
+          platform = Some(captureBody.platform),
+          ophanVisitId = None
+        ).leftMap { errorMessage =>
+          error(s"Unable to store the metadata while capturing the payment. Continuing anyway. Error: $errorMessage")
+        }
+        capture
+      }
+      .fold(error => {
+        logger.error(s"Unable to capture the payment: $error")
```
Mullefa, line 69: Are we ok logging the full error? i.e. could it contain sensitive information?

app/services/PaypalService.scala
```diff
@@ -130,17 +147,63 @@ class PaypalService(
...
+
+    val result = for {
+      payment <- asyncExecute(Payment.get(apiContext, paymentId))
+      transaction <- tryToEitherT("get payment transaction")(payment.getTransactions.asScala.head)
+      authorisation <- tryToEitherT("get transaction auth")(transaction.getRelatedResources.asScala.head.getAuthorization)
+      _ <- patchCustomId(payment, transaction) // side effect
+      r <- asyncExecute(authorisation.capture(apiContext, capture(transaction)))
+    } yield r
+
+    result transform {
```
Mullefa: This could be re-written as a flat map

app/models/PaypalApiError.scala
```diff
@@ -0,0 +1,39 @@
...
+  case object Other extends PaypalErrorType
+
+  def fromPaypalError(error: Error): PaypalErrorType = error.getName match {
+    case "INVALID_RESOURCE_ID" => NotFound
+    case "PAYMENT_ALREADY_DONE" => PaymentAlreadyDone
+    case _ => Other
+  }
+}
+
+case class PaypalApiError(
```
Mullefa, line 25: 👍

app/services/PaypalService.scala
```diff
@@ -56,13 +54,32 @@ class PaypalService(
...
-      exception.getMessage
+  private def asyncExecute[A](block: => A)(implicit tags: LoggingTags): EitherT[Future, PaypalApiError, A] = {
+
+    def detailToString(details: ErrorDetails): String =
+      s"""
+       |field: ${details.getField}
+       |issue: ${details.getIssue}
+       |""".stripMargin
+
+    def exceptionToPaypalError(exception: Throwable): PaypalApiError = exception match {
```
Mullefa: It feels like it'd be cleaner to have this logic in a static method `fromException` in the `PaypalApiError` companion object (?)

app/services/PaypalService.scala
```diff
@@ -130,17 +147,63 @@ class PaypalService(
...
+      case Right(p) if p.getState.toUpperCase != "APPROVED" =>
+        Left(PaypalApiError(s"payment returned with state: ${payment.getState}"))
       case Right(p) => Right(p)
       case Left(error) => Left(error)
     }
 
   }
 
+  def tryToEitherT[A](action: String)(block: => A)(implicit tags: LoggingTags): EitherT[Future, PaypalApiError, A] = {
+    EitherT.fromEither[Future](Either.catchNonFatal(block).leftMap { t: Throwable =>
```
Mullefa: `Future(block).attemptT` seems cleaner (and then left mapping the result); and with this implementation,`attemptT` may be a better method name (?)

app/services/PaypalService.scala
```diff
@@ -56,13 +54,32 @@ class PaypalService(
...
+          .map(detailToString)
+          .mkString(";\n")
+        PaypalApiError(PaypalErrorType.fromPaypalError(paypalException.getDetails), details)
+      case exception: Exception =>
+        error("Error while calling PayPal API", exception)
+        PaypalApiError(exception.getMessage)
     }
-  })
+
+    EitherT(Future {
```
Mullefa: `Future(block).attemptT.leftMap(exceptionToPaypalError)` seems cleaner

app/controllers/httpmodels/CaptureRequest.scala
```diff
@@ -0,0 +1,19 @@
...
+  idUser: Option[IdentityId],
+  cmp: Option[String],
+  intCmp: Option[String],
+  refererPageviewId: Option[String],
+  refererUrl: Option[String],
+  ophanPageviewId: Option[String],
+  ophanBrowserId: Option[String])
+
+object CaptureRequest {
+  implicit val jf: Format[CaptureRequest] = Json.format[CaptureRequest]
```
Mullefa: I think a fully qualified name would be preferable here e.g. to prevent ambiguity when using IntelliJ to inspect what implicits have been passed to a function

app/models/PaypalApiError.scala
```diff
@@ -0,0 +1,39 @@
...
+  errorType: PaypalErrorType,
+  message: String
+)
+
+object PaypalApiError {
+  def apply(message: String): PaypalApiError = PaypalApiError(PaypalErrorType.Other, message)
```
Mullefa: If this method was renamed to `fromString`, you'd no longer be overloading (good IMO) and also be able to create a `Writes[PaypalApiError]` using `Json.writes` instead of having to do so manually.

app/models/PaypalApiError.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  // defining the Writes manually, otherwise Play-Json gets offended by the overloaded "apply" just above
+  implicit val jf: Writes[PaypalApiError] = new Writes[PaypalApiError] {
```
Mullefa: fully qualified name (?)

app/services/PaypalService.scala
```diff
@@ -130,17 +147,63 @@ class PaypalService(
     }
   }
 
-  def executePayment(paymentId: String, payerId: String)(implicit tags: LoggingTags): EitherT[Future, String, Payment] = {
+  def executePayment(paymentId: String, payerId: String)(implicit tags: LoggingTags): EitherT[Future, PaypalApiError, Payment] = {
     val payment = new Payment().setId(paymentId)
     val paymentExecution = new PaymentExecution().setPayerId(payerId)
     asyncExecute(payment.execute(apiContext, paymentExecution)) transform  {
-      case Right(p) if p.getState.toUpperCase != "APPROVED" => Left(s"payment returned with state: ${payment.getState}")
+      case Right(p) if p.getState.toUpperCase != "APPROVED" =>
```
Mullefa: This could be re-written using flat map

app/controllers/PaypalController.scala
```diff
@@ -77,8 +106,8 @@ class PaypalController(paymentServices: PaymentServices, corsConfig: CorsConfig,
         ophanVisitId = ophanVisitId
       )
 
-    def notOkResult(message: String): Result = {
-      error(s"Error executing PayPal payment for request id: ${request.id} \n\t error message: $message")
+    def notOkResult(paypalError: PaypalApiError): Result = {
```
Mullefa, line 109: 👍

Mullefa (commented):
Mainly commenting on syntax in this review. Is it ok to have a quick chat after lunch just to understand the motivation behind this pull request ?

app/controllers/PaypalController.scala
alexduf: Yep, this is of the type `PaypalApiError` that I define in this PR

app/services/PaypalService.scala
alexduf: Good shout!

app/services/PaypalService.scala
alexduf: What a nifty function, thanks!

app/models/PaypalApiError.scala
alexduf: Yeah I know, but I was hesitating between easy use of the PaypalApiError class, or writing a two field Writer. I'll use the fromString function instead

app/services/PaypalService.scala
Mullefa: \* subflatmap 😅

Mullefa (approved):

PR #329, opened 2017-08-23
reposition stripe payment complete reporting and add metric to track unhandled stripe failures

michaelwmcnamara:
This pull request repositions payment complete reporting in Stripe.controller so that it fires at the very last moment before user is redirected to thank-you page.

This is because there have been a number of false alarms as a result of payment being successful, but user not being shown the thank-you page. This calls the use of thank-you page as a reliable indicator of a successful payment into question and I have changed the librato metrics to instead alert on the Stripe Payment Success metric insted.
To minimise the chance of this alert not being triggered in the event of an unsuccessful payment, I have moved it into the redirect logic called at the very end of the "pay" function.

This pull request also adds a new metric to capture any Stripe payment failures that are not a direct result of Stripe responding that it does not accept thee payments. The idea of this metric is to capture edge cases such as timeouts.

### DIFF
--- a/app/controllers/StripeController.scala
+++ b/app/controllers/StripeController.scala
@@ -83,6 +83,8 @@ class StripeController(paymentServices: PaymentServices, stripeConfig: Config, c
       cloudWatchMetrics.logPaymentSuccessRedirected(PaymentProvider.Stripe, request.platform)
       mobileRedirectUrl(contributionAmount)
     } else {
+      info(s"Stripe payment successful for request id: ${request.id}, from platform ${request.platform}")
+      cloudWatchMetrics.logPaymentSuccess(PaymentProvider.Stripe, request.platform)
       routes.Contributions.thanks(countryGroup).url
     }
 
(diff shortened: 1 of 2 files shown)


app/monitoring/CloudWatchMetrics.scala
```diff
@@ -67,6 +67,10 @@ class CloudWatchMetrics(cloudWatchClient: AmazonCloudWatchAsync) extends TagAwar
     put("payment-failure", paymentProvider, platform)
   }
 
+  def logUnhandledPaymentFailure(paymentProvider: PaymentProvider, platform: String): Unit = {
+    put("payment-failure", paymentProvider, platform)
```
desbo: Should this be called `unhandled-payment-failure` ?

michaelwmcnamara: omg yes it should!!!

michaelwmcnamara: Latest commit fixes this.

app/controllers/StripeController.scala
```diff
@@ -130,8 +132,6 @@ class StripeController(paymentServices: PaymentServices, stripeConfig: Config, c
     }
 
     createCharge.map { charge =>
-      info(s"Stripe payment successful for request id: ${request.id}, from platform ${request.platform}")
```
desbo: I think this change should be reverted, since it doesn't look like moving this into the `thankYouUri` call will make any difference. In both cases, the calls would be executed before the user gets the redirect response.

It'd be good to keep the `thankYouUri` function as something that purely generates a URI without any unrelated side effects - what about a new function to take care of the logging stuff?

michaelwmcnamara: These are good points! - this is fixed in latest commit.

desbo (approved):

PR #330, opened 2017-08-24
Provide Payment object instead of payment ID when storing PayPal metadata

desbo:
This is to prevent what seems to be an unneccessary call to the PayPal API, and will allow me to avoid any 3rd-party dependencies when sending Acquisition events (which will be in a future PR).

At the moment, we provide the payment ID to the `executePayment` call in the controller, which gives us back a `Payment`. We then pass the payment ID to the `storeMetaData` function which goes on to call the PayPal API to retrieve the Payment object. This seems redundant since we can instead provide the Payment from `executePayment` to `storeMetaData`, which is what this PR changes.


alexduf (approved):
Makes sense! 👍

PR #332, opened 2017-08-30
logging improvements to use play session id to track payments

michaelwmcnamara:
This is a change to the logging - essentially it replaces request id with the play session id.
In addition, the id is recorded as "payment_session" field in the play session cookie. This is needed because both stripe and paypal payments have a new session id and request id for the thank-you and post payment pages, so it is very difficult to follow a payment to completion with any confidence in the current logging setup.
Note - I have kept payment hooks using the request id - this is because I don't believe that they will have a play session cookie associated with them.

Testing - I have tested both paypal and stripe flows and both work ok locally.

### DIFF
--- a/app/actions/CommonActions.scala
+++ b/app/actions/CommonActions.scala
@@ -62,27 +62,20 @@ object CommonActions {
       case _ => _platform
     }
 
-    def playId: Option[String] = {
+    def playSessionId: Option[String] = {
       for {
         playSessionCookie <- request.cookies.get("PLAY_SESSION")
         value <- Option(playSessionCookie.value)
       } yield {
-        value.take(value.indexOf("-"))
+        value.takeWhile( _ != '-')
       }
     }
 
-    def initialSessionId: String = {playId.getOrElse("")}
+    def initialSessionId: String = playSessionId.getOrElse("unknown_contributions_session")
 
     def sessionId: String = {
-      val payment_session = request.session.get("payment_session")
-      payment_session match {
-        case Some(id) =>
-          if(id.nonEmpty)
-            id
-          else
-            initialSessionId
-        case None => initialSessionId
-          }
+      val contributionSession = request.session.get("contributions_session")
+      contributionSession.getOrElse(initialSessionId)
     }
 
     def isAndroid: Boolean = platform == "android"
--- a/app/actions/CommonActions.scala
+++ b/app/actions/CommonActions.scala
@@ -74,8 +74,7 @@ object CommonActions {
     def initialSessionId: String = playSessionId.getOrElse("unknown_contributions_session")
 
     def sessionId: String = {
-      val contributionSession = request.session.get("contributions_session")
-      contributionSession.getOrElse(initialSessionId)
+      request.session.get("contributions_session").getOrElse(initialSessionId)
     }
 
     def isAndroid: Boolean = platform == "android"
--- a/app/actions/CommonActions.scala
+++ b/app/actions/CommonActions.scala
@@ -62,6 +62,29 @@ object CommonActions {
       case _ => _platform
     }
 
+    def playId: Option[String] = {
+      for {
+        playSessionCookie <- request.cookies.get("PLAY_SESSION")
+        value <- Option(playSessionCookie.value)
+      } yield {
+        value.take(value.indexOf("-"))
+      }
+    }
+
+    def initialSessionId: String = {playId.getOrElse("")}
+
+    def sessionId: String = {
+      val payment_session = request.session.get("payment_session")
+      payment_session match {
+        case Some(id) =>
+          if(id.nonEmpty)
+            id
+          else
+            initialSessionId
+        case None => initialSessionId
+          }
+    }
+
     def isAndroid: Boolean = platform == "android"
     def isIos: Boolean = platform == "ios"
     def isMobile: Boolean = isAndroid || isIos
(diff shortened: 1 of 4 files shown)


app/controllers/PaypalController.scala
```diff
@@ -124,15 +124,15 @@ class PaypalController(paymentServices: PaymentServices, corsConfig: CorsConfig,
         case Accepts.Html() =>
           val amount: Option[ContributionAmount] = paypalService.paymentAmount(payment)
           val email = payment.getPayer.getPayerInfo.getEmail
-          val session = List("email" -> email, PaymentProvider.sessionKey -> PaymentProvider.Paypal.entryName) ++ amount.map("amount" -> _.show)
+          val session = List("email" -> email,PaymentProvider.sessionKey -> PaymentProvider.Paypal.entryName) ++ amount.map("amount" -> _.show)
```
Mullefa: Can you re-add the space to remove this diff?

app/actions/CommonActions.scala
```diff
@@ -62,6 +62,29 @@ object CommonActions {
...
+    def playId: Option[String] = {
+      for {
+        playSessionCookie <- request.cookies.get("PLAY_SESSION")
+        value <- Option(playSessionCookie.value)
+      } yield {
+        value.take(value.indexOf("-"))
+      }
+    }
+
+    def initialSessionId: String = {playId.getOrElse("")}
```
Mullefa: I think there should be a more informative default value e.g. `unknown_play_session`

Mullefa: `{}` not needed around expression

app/actions/CommonActions.scala
```diff
@@ -62,6 +62,29 @@ object CommonActions {
       case _ => _platform
     }
 
+    def playId: Option[String] = {
```
Mullefa: `{}` not needed around expression

Mullefa: I think a better name would be `playSessionId`

app/actions/CommonActions.scala
```diff
@@ -62,6 +62,29 @@ object CommonActions {
...
+      for {
+        playSessionCookie <- request.cookies.get("PLAY_SESSION")
+        value <- Option(playSessionCookie.value)
+      } yield {
+        value.take(value.indexOf("-"))
```
Mullefa: I think `value.takeWhile(_ != '-')` would be clearer. Also, why are we only taking the characters up to the first `-`?

app/controllers/Contributions.scala
```diff
@@ -99,7 +99,7 @@ class Contributions(paymentServices: PaymentServices, addToken: CSRFAddToken, cl
         CSRF.getToken.map(_.value),
         request.isAllocated(Test.landingPageTest, "with-copy"),
         disableStripe
-      ))
+      )).addingToSession("payment_session" -> request.sessionId)
```
Mullefa: ~~What does adding the `payment_session` value to the session give us? As opposed to just using `initialSessionId`?~~

app/actions/CommonActions.scala
```diff
@@ -62,6 +62,29 @@ object CommonActions {
...
+      }
+    }
+
+    def initialSessionId: String = {playId.getOrElse("")}
+
+    def sessionId: String = {
```
Mullefa: ~~If we don't need to set the `payment_session` value (see comment below), then this method can be removed completely, and `initialSessionId` renamed to `sessionId`.~~

Mullefa (commented):
In addition to the comments, I think we can refer to the session id in all the log message as `session id` as opposed to `payment_session`

app/controllers/Contributions.scala
Mullefa: I think something like `contributions_session` would be a more informative name.

app/actions/CommonActions.scala
```diff
@@ -62,6 +62,29 @@ object CommonActions {
...
+      val payment_session = request.session.get("payment_session")
```
Mullefa: camelCase please

app/actions/CommonActions.scala
```diff
@@ -62,6 +62,29 @@ object CommonActions {
...
+      payment_session match {
+        case Some(id) =>
```
Mullefa: If `initialSessionId` returned a non-empty string for missing values, this case statement could be rewritten as: `paymentSession.getOrElse(initialSessionId)`

app/actions/CommonActions.scala
michaelwmcnamara: Agree takeWhile(_ != '-') would be more elegant and clearer.

Reason we are only taking characters up to the first - is that - as far as I can tell - the id field isn't directly accessible as a field through Play. It is annoyingly part of the cookie value, but not part of the .data fields that follow it. 
Therefore to get the Play session id you have to access the cookie value then strip out all the other fields that have been added to the session cookie. The delimiter between the id field and the data fields is the '-' character. 
Main reason this is necessary is that between payment completion and the post-payment/thank-you-page stages of the payment process, the request id and play session id are overwritten. Therefore on display of the home page I append the Play session key to the play session as the "payment_session" field. This means it persists into the pos-payment page and thank-you page and can be used to tie these together.

If I was to use the full value of the play session cookie for this, there would be a lot of fields duplicated as part of the "payment_session" field which could lead to confusion, as well as being too verbose.

app/controllers/Contributions.scala
michaelwmcnamara: changed all references to "payment_session" to read "contributions_session" in next commit

app/actions/CommonActions.scala
michaelwmcnamara: *shameFace* fixed in next commit

app/actions/CommonActions.scala
michaelwmcnamara: Good idea! Fixed in next commit.

Mullefa (approved):

PR #333, opened 2017-08-31
Stripe meta data update

Mullefa:
Country code and country subdivision code added to the the Stripe meta data. This allows contributions in the US to be audited by state. See corresponding [code](https://manage.fastly.com/configure/services/14OHcjf7ueXm1ilie2wY00/versions/13/contents) and [production](https://manage.fastly.com/configure/services/3eAImrcMhkiYNBpDzEIpbx/versions/10/contents) configurations in Fastly.

Here is a fragment from the Stripe Dashboard. The record was created by using a Stripe test card on CODE.

stripe-meta-data

Have you gone through the contributions flow for all test variants ? __Yes, tested on CODE with Stripe and Paypal test accounts__

### DIFF
--- a/app/utils/FastlyUtils.scala
+++ b/app/utils/FastlyUtils.scala
@@ -3,10 +3,17 @@ package utils
 import com.gu.i18n.CountryGroup
 import play.api.mvc.Request
 
-object RequestCountry {
-  implicit class RequestWithFastlyCountry(r: Request[_]) {
-    def getFastlyCountry = r.headers.get("X-GU-GeoIP-Country-Code").flatMap(CountryGroup.byFastlyCountryCode)
+object FastlyUtils {
+
+  implicit class FastlyRequest(r: Request[_]) {
+
+    def getFastlyCountryCode: Option[String] = r.headers.get("X-GU-GeoIP-Country-Code")
+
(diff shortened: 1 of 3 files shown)


app/utils/FastlyUtils.scala
```diff
@@ -3,10 +3,17 @@ package utils
 import com.gu.i18n.CountryGroup
 import play.api.mvc.Request
 
-object RequestCountry {
-  implicit class RequestWithFastlyCountry(r: Request[_]) {
-    def getFastlyCountry = r.headers.get("X-GU-GeoIP-Country-Code").flatMap(CountryGroup.byFastlyCountryCode)
+object FastlyUtils {
+
+  implicit class FastlyRequest(r: Request[_]) {
```
desbo, line 8: 👍

desbo (approved):
Nice work 🇺🇸

app/utils/FastlyUtils.scala
```diff
@@ -3,10 +3,17 @@ package utils
...
+
+    def getFastlyCountryCode: Option[String] = r.headers.get("X-GU-GeoIP-Country-Code")
```
desbo, line 10: Interesting to note that the `X-` prefix we use for this one is deprecated: https://tools.ietf.org/html/rfc6648 - wouldn't say it's worth changing though.

PR #334, opened 2017-09-05
Allow test username to be passed as a query parameter

davidfurey:
This will make it easier to support test users in support-frontend without needing to add CSRF complexity associated with cookies

No


jranks123 (approved):
Seems sensible!

Issue #336, opened 2017-09-07
Paypal capture payments logic


Mullefa:
When [capturing payments](https://github.com/guardian/contributions-frontend/blob/master/app/controllers/PaypalController.scala#L51), if we fail to get the payment, the client will get an internal server error which interpreted to mean the payment has not been made, even though it has been.

PR #337, opened 2017-09-07
Use throwables to encode errors rather than strings in DB access

Mullefa:
This pull request uses throwables to encode errors rather than strings for data base access. This means that it is easier to combine such return types with the custom `PaypalAPIError` (introduced in #325) in a monadic context. This composition will be used when fixing issue #336.

### DIFF
--- a/app/controllers/StripeController.scala
+++ b/app/controllers/StripeController.scala
@@ -113,7 +113,7 @@ class StripeController(paymentServices: PaymentServices, stripeConfig: Config, c
       )
     }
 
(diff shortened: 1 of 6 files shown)


alexduf:
I would discourage you to add the `Throwable` trait to any object, unless you really plan on throwing it at some point by doing `throw new MyObject`, which is often discouraged in scala.

If what you want is just a marker trait, maybe define something like `ContributionError`.

Mullefa:
Thanks Alex - as discussed I'll make an attempt in this pull request to encode the errors properly, but will prioritise #336 over this. To be continued...

PR #338, opened 2017-09-07
Fix for #336

Mullefa:
Fix of issue #336


Mullefa:
@michaelwmcnamara -  @alexduf is going to review this, but can you check you are ok with the errors that are logged?

alexduf (approved):
Nice, thanks @Mullefa !

PR #346, opened 2017-09-18
Contributions session id added to warning

Mullefa:
Contributions session id added to warning log.


michaelwmcnamara (approved):
Make it so!

PR #347, opened 2017-09-19
Fix custom session id

Mullefa:
cc @guardian/contributions 

Currently we are using the signature of the Play session cookie as the session id - not good for providing a unique id per user session.

The solution is to generate a unique id and then persist it across requests using the Play session cookie via an action builder. 

For simplicity, I've added to (and renamed) the pre-existing `AbTestAction` builder, rather than create a new one and compose them.

_Have you gone through the contributions flow for all test variants ?_ __Yes, on CODE__

### DIFF
--- a/app/abtests/Test.scala
+++ b/app/abtests/Test.scala
@@ -1,6 +1,6 @@
 package abtests
 
-import actions.CommonActions.ABTestRequest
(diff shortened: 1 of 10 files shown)


michaelwmcnamara:
Looks good!

michaelwmcnamara (approved):

PR #349, opened 2017-09-20
Handle error thrown when parsing invalid JSON

Mullefa:
I didn't realise that `Json.parse()` didn't encode the error in the result. As noted by @joelochlann , this resulted in an error page (screenshot below) if the value of `acquisitionData` in the query string was not valid JSON (__although no user saw this as we currently aren't using this query string field__).

error

This pull request fixes this issue.


joelochlann (approved):
Nice one!

PR #350, opened 2017-09-20
Add support for regional Stripe configurations

desbo:
This adds a new layer to the `stripe.keys` config section that groups keys by Fastly country code and updates the `PaymentServices` logic accordingly. 

There's a new release of `membership-common` on the way which will require us to pass in a `Country` when creating `StripeApiConfig`, but that can be done in a separate PR when we update the dependency. 

Note that this will require a change to the configs in S3, which should be done immediately before deployment.

### DIFF
--- a/app/controllers/StripeController.scala
+++ b/app/controllers/StripeController.scala
@@ -44,15 +44,15 @@ class StripeController(paymentServices: PaymentServices, stripeConfig: Config, c
 
     val form = request.body
 
-    val stripe = paymentServices.stripeServiceFor(form.name, request)
-    val idUser = IdentityId.fromRequest(request) orElse form.idUser
-
     val countryGroup = form.currency match {
       case USD => US
       case AUD => Australia
       case EUR => Europe
       case _ => UK
     }
+
+    val stripe = paymentServices.stripeServiceFor(form.name, Some(countryGroup))
+    val idUser = IdentityId.fromRequest(request) orElse form.idUser
 
     val contributionId = ContributionId.random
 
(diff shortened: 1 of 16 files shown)


conf/PROD.public.conf
```diff
@@ -12,12 +12,26 @@ stripe {
...
-        LIVE {
-            public = "pk_live_auSwLB4KBzbN3JOUVHvKMe6f"
+
+        au {
+            TEST {
+                public = "pk_test_I1ts3iShWrjssTavL0b7QXQ6"
+            }
+
+            LIVE {
+                public = "CHANGEME"
```
desbo: waiting on S. Bibb for this

app/services/PaymentServices.scala
```diff
@@ -23,18 +24,28 @@ class PaymentServices(
   actorSystem: ActorSystem
 )(implicit ec: ExecutionContext) {
 
-  val stripeServices: Map[PaymentMode, StripeService] = {
+  import utils.ConfigUtils._
+  import utils.FastlyUtils._
+
+  def stripeServices(maybeCountryGroup: Option[CountryGroup]): Map[PaymentMode, StripeService] = {
```
Mullefa: as discussed, preferable not to create a new instance of the map each time this function is invoked

app/controllers/StripeController.scala
```diff
@@ -193,7 +193,9 @@ class StripeController(paymentServices: PaymentServices, stripeConfig: Config, c
       }
 
       withParsedStripeHook(request.body) { stripeHook =>
-        val stripeService = paymentServices.stripeServices(stripeHook.mode)
+        val countryGroup: Option[CountryGroup] = CountryGroup.byFastlyCountryCode(stripeHook.fastlyCountryCode)
```
dominickendrick, line 196: Just to check, do we only want to switch to AUS on the country code ? Should we check for currency in AUD as well ? Can people pay in AUD from other countries ??

desbo: @dominickendrick the only way to pay in AUD is by choosing that country group, and we'll use the right key in that case. The reason I'm inferring it from the header here is because this comes from the Stripe hook where the country group is stored in the metadata.

dominickendrick: Ah ok, makes sense : )

dominickendrick (approved):

PR #353, opened 2017-09-25
Ophan service instance with correct handling of test and default requests

Mullefa:
cc @guardian/contributions 

The current Ophan instance used by the app is determined by inspecting the mode the Play app is running in - if in production mode, then the all acquisition events are sent to the production instance of Ophan, else an instance is created which fires events to a test endpoint (if specified in the config).

This implementation has some minor issues (which this pull request resolves):

- contributions frontend on CODE is running in production mode, meaning all acquisitions on CODE will be sent to Ophan
- test requests on the live version of the website will be sent to Ophan

In addition to this pull request, the `ophan.default` field in the private configuration for code will have to be overridden to `TEST`. This mirrors the pattern of how payment and database instances are determined for each request. 

### DIFF
--- a/conf/DEV.public.conf
+++ b/conf/DEV.public.conf
@@ -82,5 +82,14 @@ contexts {
 }
 
 ophan {
-    isProd = false
+
+
+    default = TEST
+    testing = TEST
+
+    # Use to configure which endpoint the TEST Ophan service should submit acquisition events to.
+    # If testEndpoint is not specified then the TEST Ophan service will still attempt to build acquisition events,
+    # but will not submit them to an endpoint.
+
+    # testEndpoint = "http://localhost:9000"
 }
--- a/conf/DEV.public.conf
+++ b/conf/DEV.public.conf
@@ -81,4 +81,6 @@ contexts {
     }
 }
 
-ophan.endpoint = "http://localhost:9000"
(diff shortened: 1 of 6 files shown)


conf/DEV.public.conf
```diff
@@ -81,4 +81,15 @@ contexts {
     }
 }
 
-ophan.endpoint = "http://localhost:9000"
+ophan {
+
+
+    default = TEST
+    testing = TEST
```
joelochlann, line 88: Maybe we could name this so it's absolutely clear what it does? `whenTestUserLoggedIn` or something like that?

Mullefa: I agree, but there's already a precedent in the config file which I don't really want to change in this pull request e.g. it would mean updating the PROD and DEV public and private configs.

joelochlann (approved):
Great stuff! It did take a bit of reading for me to get the logic though, partly because of the ambiguity of the word `test` and `testing`. Maybe we could consider making it explicit in the variable names when "testing" is exactly equivalent to "logged in as test user"

Mullefa:
@joelochlann I agree - was confused too. I think an overhaul of terminology would be beneficial if/when there's a bit of downtime.

PR #359, opened 2017-10-11
/stripe/pay endpoint update

Mullefa:
Previously, the field `abTest` in the body of the request to `/stripe/pay` was used to record the AB test the user was in on the referring page; the tests they were in on the contributions website was inferred from a cookie. However, when `/stripe/pay` is called from support frontend, the AB tests the user was in on the Supporter website can not be included in the request.

This pull request updates the `/stripe/pay` endpoint so that the fields `refererAbTest` and `nativeAbTests` can be included in the body of the request, and sets a precedent that they should be used to record the (optional) AB test the user was in on the referring page, and the AB tests they were in on the native website respectively.

See [#270](https://github.com/guardian/support-frontend/pull/270) on support-frontend which uses this updated endpoint.

Here is an example of the data that is getting posted to the `/stripe/pay` endpoint from the contributions website:

stripe_pay_request_body

Have you gone through the contributions flow for all test variants ? __Yes, locally, on code and from Support frontend on code__

### DIFF
--- a/app/controllers/forms/ContributionRequest.scala
+++ b/app/controllers/forms/ContributionRequest.scala
@@ -28,7 +28,9 @@ case class ContributionRequest(
   componentId: Option[String],
   componentType: Option[ComponentType],
   source: Option[AcquisitionSource],
-  abTest: Option[AbTest]
+  abTest: Option[AbTest], // Deprecated
+  refererAbTest: Option[AbTest],
+  nativeAbTests: Option[Set[AbTest]]
 )
 
 object ContributionRequest {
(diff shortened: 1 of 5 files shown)


app/controllers/forms/ContributionRequest.scala
```diff
@@ -28,7 +28,9 @@ case class ContributionRequest(
   componentId: Option[String],
   componentType: Option[ComponentType],
   source: Option[AcquisitionSource],
-  abTest: Option[AbTest]
+  abTest: Option[AbTest], // Deprecated
+  refererAbTest: Option[AbTest],
+  nativeAbTests: Option[Set[AbTest]]
```
desbo, line 33: would a better name be `contributionsAbTests`?

Mullefa: No, because this field is going to be used by S&C as well. Open to another name though...

desbo (approved):

PR #363, opened 2017-10-13
URL do not exist

kant:


desbo:
Thanks @kant  - the correct URL is actually https://github.com/guardian/riff-raff/blob/master/riff-raff/public/docs/howto/fix-a-failed-deploy.md, would you mind changing it to that?

PR #369, opened 2017-11-03
Adding sbt-dependency-graph plugin

michaelwmcnamara:
I am evaluating snyk.io to monitor dependencies vulnerabilities.
Snyk.io depends on the sbt-dependency-graph plugin

It is necessary to use CLI setup as the default configuration for Snyk (which goes directly to the public repository) does not pick up dev dependencies.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -12,3 +12,6 @@ addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.6.1")
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.2")
 
 addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "0.9.7")
+
+// dependency tracker plugin - needed for snyk cli integration
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.8.2")
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -14,4 +14,4 @@ addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.2")
 addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "0.9.7")
 
 // dependency tracker plugin - needed for snyk cli integration
-addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.8.2")
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.0")


michaelwmcnamara:
Equivalent step was taken by @TBonnin  when starting to use snyk on frontend project:
https://github.com/guardian/frontend/pull/17955

desbo:
I noticed support-frontend use this as well, but they don't have the dependency graph plugin - I'm not necessarily against adding it but could be worth seeing how theirs is set up as a comparison to frontend.

TBonnin:
In theory snyk can read and evaluate dependencies if they are specified in `build.sbt` (which seems to be the case for this project. Edit: scratch that I was wrong)
Using the snyk CLI (which require the `sbt-dependency-graph` plugin) is necessary when your sbt settings are more exotic (like frontend where dependencies are defined in a separate file)

michaelwmcnamara:
Hi @desbo I'm in the process of setting up support-frontend with @Ap0c . Currently its is trying to pull directly from git. 
Basically, using the Github interface, Snyk can only see dependencies if they are in your build.sbt file or in you package.json file. In order to be picked up, they must be listed directly in libraryDependencies for build.sbt (no references to a val such as we currently use).
In addition it cannot pick up dev-dependencies in the packages.json file.

Using the Snyk CLI gives you more control - you can specify to include dev-dependencies in packages.json for instance - it also uses this plugin to search for and gather dependencies that are defined outside of libraryDependencies.

michaelwmcnamara:
In a nutshell, the alternative to adding this package would be to list all our sbt dependencies under libraryDependencies in build.sbt and combine our dependencies in package.json into the one list rather than dividing it between prod dependencies and dev-dependencies.

Ap0c:
> combine our dependencies in package.json into the one list

As @michaelwmcnamara said, for the client-side specifically, to get the GitHub integration working in membership-frontend I moved all `devDependencies` to the `dependencies` section of `package.json` in [this PR](https://github.com/guardian/membership-frontend/pull/1742).

If this is not something you want to do for client-side dependencies then you need to use the CLI and pass the `--dev` flag, which picks up both `dependencies` and `devDependencies`.

project/plugins.sbt
```diff
@@ -12,3 +12,6 @@ addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.6.1")
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.2")
 
 addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "0.9.7")
+
+// dependency tracker plugin - needed for snyk cli integration
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.8.2")
```
desbo: just a minor thing, but you could use the latest version - `0.9.0`

michaelwmcnamara: @desbo  Good spot! This should be remedied now.

desbo (approved):

PR #374, opened 2017-11-17
Log payment amounts

michaelwmcnamara:
This change adds payment amount information to logging for payment success and - where possible - payment failure.

This is to make it easier to manually check the amounts of payments are as expected and to aid future investigations where users are unclear on how much they have paid.

### DIFF
--- a/app/controllers/PaypalController.scala
+++ b/app/controllers/PaypalController.scala
@@ -168,7 +168,7 @@ class PaypalController(paymentServices: PaymentServices, corsConfig: CorsConfig,
           } else {
             routes.Contributions.postPayment(countryGroup).url
           }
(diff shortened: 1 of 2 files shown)


dominickendrick (approved):
This looks good, my only question is will putting the amount charged into the logs have any GDPR impact ? Is this data we want in our logs ?

michaelwmcnamara:
I think its ok. The only danger would be in being able to tie the amount to the users session id. These identifiers are randomly generated and stored on the play session cookie. They are used so we can follow the payments from start to finish. (the interleaving of multiple transactions makes it impossible to do this otherwise). 
However this identifier only lasts as long as a browser session and will expire after they close their browser. Next time they visit a new and different session id will be generated.
Any association between that payment and the user is a fleeting one.

michaelwmcnamara:
Merging this for now as I need the logs to be able to debug a potential payment issue.
I will revert if it turns out this is a GDPR issue after all.

PR #375, opened 2017-11-30
Adding a configuration file for CODE

svillafe:
In code we need to point to the support thank you page in order to not mess up with tracking.


svillafe:
Tested in CODE and it is pointing to the right thank you page (after updating the private conf).

desbo (approved):

PR #376, opened 2017-11-30
Set the correct stage in the Code conf file

rupertbates:
@desbo I noticed that the stage was still set to PROD in the CODE.conf file so I changed it, but @svillafe told me to talk to you about it.


desbo (approved):
Thanks - I think the only difference it'll make is reporting `CODE` in some metrics/logs, but worth having.

PR #383, opened 2017-12-19
Tweak Header Brandbar And Updated Gruntfile Mapping

Ap0c:
Switched the Gruntfile to using a map for CSS assets instead of an array. Also tweaked the styling on the header and brandbar.


assets/stylesheets/garnett.scss
```diff
@@ -0,0 +1,68 @@
+// ----- Main Stylesheets ----- //
+
+@import './style';
+
+
+// ----- Garnett Styles ----- //
+
+.container-global.container-global--giraffe {
```
joelochlann, line 8: Generally in BEM you would avoid nesting and selectors with multiple classes. Though I imagine in this case you're constrained by having to override all the violations of BEM style in other parts of the CSS...

joelochlann (approved):

Ap0c:
Thanks @davidfurey and @joelochlann!

PR #384, opened 2017-12-22
Remove Ophan call in payment flow

joelochlann:
This fixes the payment flow in IE11

In IE11, all Ophan requests are aborted by the browser. This means the promise which wraps `ophan.record` never resolves so the Stripe & PayPal flows never move on to the next step.

Since we don't need the complete information, we've just removed these calls

### DIFF
--- a/assets/javascripts/src/actions.es6
+++ b/assets/javascripts/src/actions.es6
@@ -3,7 +3,7 @@ import 'whatwg-fetch';
 import store from 'src/store';
 import { urls } from 'src/constants';
 import { trackCheckout, trackPayment } from 'src/modules/analytics/ga';
-import { completeTests, inStripeCheckoutTest } from 'src/modules/abTests';
+import { inStripeCheckoutTest } from 'src/modules/abTests';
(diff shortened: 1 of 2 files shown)


jranks123 (approved):
Such a lovely PR to see off the year :1st_place_medal:

mchv:
@joelochlann do you know why  in IE11, all Ophan requests are aborted by the browser? do you think we may have the same problem on the frontend?

joelochlann:
@mchv sorry for late reply - we do actually have the same problem in frontend, I checked. I can reproduce in IE11 on Windows 7 and Windows 8.1, but on Windows 10 it seems to be fine.

The question is whether the request does actually get through to Ophan or not. Even when the request shows as "Aborted" in the Network tab, I can still see the response header (`204 No Content`) if I click on the "Detail" tab for that request. I wonder if certain versions of IE11 handle the request strangely if it returns with a 204 code.

This issue seems to be the similar: https://forum.matomo.org/t/internet-explorer-tracking-request-aborted/18177

PR #387, opened 2018-01-09
Solicit email consents using new Identity Consent API

jranks123:
This PR implements the new marketing email consent process for contribute.theguardian.com.

As part of the change, we are asking for consent AFTER payment for both Stripe and Paypal payments. This is because the new copy (which is not yet in this PR) is considered too long to place in the flow for fear that it may lower the conversion rate. 

It also defaults the box to unchecked. 

Here is a screenshot of the consent box 

contrib

If you opt in, you get this email:
contribemail

Which takes you to this page:
thabkny


app/controllers/IdentityController.scala
```diff
@@ -0,0 +1,33 @@
...
+      case Some(email) => identityService.updateMarketingOptIn(email, marketingOptIn)
+      case None => Future.successful(error("email not found in session while trying to update marketing opt in"))
+    }
+
+    val url = request.session.get("amount").flatMap(ContributionAmount.apply)
+      .filter(_ => request.isAndroid)
+      .map(mobileRedirectUrl)
+      .getOrElse(routes.Contributions.thanks(countryGroup).url)
+
+    contributor.map { _ =>
```
desbo: I don't think you need to use `map` here, or even create `contributor`. Just evaluating the `match` statement will be enough.

app/services/IdentityService.scala
```diff
@@ -72,4 +85,10 @@ class IdentityService(wsClient: WSClient, config: Config)(implicit ec: Execution
           error("Error accessing Identity API", err)
           Autofill.empty
       }
+
+  def updateMarketingOptIn(email: String, marketingOptInt: Boolean)(implicit tags: LoggingTags): Future[Boolean] = {
+    if (marketingOptInt)
```
desbo: should be `marketingOptIn`

conf/routes
```diff
@@ -20,7 +20,7 @@ OPTIONS        /paypal/auth                     controllers.PaypalController.aut
 POST           /paypal/capture                  controllers.PaypalController.capturePayment
 GET            /paypal/:countryGroup/execute    controllers.PaypalController.executePayment(countryGroup: CountryGroup, paymentId, token, PayerID, CMP: Option[String], INTCMP: Option[String], refererPageviewId: Option[String], refererUrl: Option[String],pvid: Option[String], bid: Option[String], ophanVisitId: Option[String], componentId: Option[String], componentType: Option[ComponentType], source: Option[AcquisitionSource], refererAbTest: Option[AbTest], nativeAbTests: Option[Set[AbTest]], supportRedirect: Option[Boolean])
 POST           /paypal/hook                     controllers.PaypalController.hook
-POST           /:countryGroup/update-metadata   controllers.PaypalController.updateMetadata(countryGroup:CountryGroup)
+POST           /:countryGroup/update-marketing  controllers.IdentityController.updateMarketingPreferences(countryGroup:CountryGroup)
```
davidfurey, line 23: Keep both routes for now to avoid breaking cached JS

app/services/StripeService.scala
```diff
@@ -100,6 +100,8 @@ class StripeService(
     idUser.foreach { id =>
       identityService.updateMarketingPreferences(id, marketing)
     }
+    if (marketing)
```
joelochlann: we can remove this, plus lines 100 to 102, plus `marketing` param from this function

app/services/PaypalService.scala
```diff
@@ -289,6 +289,8 @@ class PaypalService(
     idUser.map { id =>
       identityService.updateMarketingPreferences(id, marketingOptInt)
     }
+    if (marketingOptInt)
```
joelochlann: you can remove this whole method

app/services/IdentityService.scala
```diff
@@ -72,4 +85,10 @@ class IdentityService(wsClient: WSClient, config: Config)(implicit ec: Execution
           error("Error accessing Identity API", err)
           Autofill.empty
       }
+
+  def updateMarketingOptIn(email: String, marketingOptIn: Boolean)(implicit tags: LoggingTags): Future[Boolean] = {
```
joelochlann, line 78: you can delete `updateMarketingPreferences` from this class

app/services/PaypalService.scala
joelochlann: sorry - I mean, you can get rid of this whole method!

joelochlann (changes requested):
As mentioned in individual comments, you can get rid of some redundant code

Also, I think the marketing preferences should also be removed from anywhere it's being sent as metadata to Stripe & PayPal - I imagine we're probably not GDPR-compliant if that doesn't happen (because the user has no way of updating that metadata)

app/services/IdentityService.scala
jranks123: I can't delete this as this one as it is called from the Identity Controller, which is somewhere we want to call this from

conf/routes
jranks123: Done

app/services/IdentityService.scala
jranks123: done

app/services/StripeService.scala
```diff
@@ -95,21 +95,16 @@ class StripeService(
     idUser: Option[IdentityId],
     marketing: Boolean)
```
joelochlann, line 96: you can delete this param - it's not used now

app/controllers/forms/ContributionRequest.scala
```diff
@@ -47,4 +49,15 @@ object ContributionRequest {
   implicit val contributionRequestReads: Reads[ContributionRequest] = Json.reads[ContributionRequest]
 }
 
+case class MetadataUpdate(marketingOptIn: Boolean)
+
+object MetadataUpdate {
```
joelochlann, line 54: Could we replace `metadata` everywhere here with `marketingOptIn` since that's all it does - the word metadata is a bit overloaded and keeps making me wonder whether this is being used to update Stripe/PayPal metadata - which it's not

joelochlann: in fact - I'd be inclined to just put this inside `IdentityController` since that's the only place it's used and ever will be used

joelochlann (approved):

joelochlann:
Nice work Jonny, especially with the tight deadline.

As discussed, we just need to make sure that (in a subsequent PR) we remove `marketing` as a parameter on `ContributionRequest` before merging the changes to the support site - otherwise when we stop supplying it the JSON parsing will break

PR #395, opened 2018-01-12
Stop sending marketing preferences to Stripe and Postgres

jranks123:
Consent to receiving marketing is now taken and stored through Identity. This means that we should not be sending marketing preference to Stripe, Paypal or Postgres.


joelochlann (approved):
Looks good!

PR #398, opened 2018-01-15
Garnett

Ap0c:
## Why are you doing this?

To launch Garnett.

## Changes

- Apply the new logo.
- Apply the new roundel.
- Turn off the Garnett flag in Gruntfile, serve Garnett by default.


davidfurey (approved):
🐦 👍

PR #403, opened 2018-01-17
add an eventually and a timeout to the StoreMetaData checker

michaelwmcnamara:
Attempt to resolve acceptance test failures issue described in https://trello.com/c/v1NXxFH1/35-intermittent-test-failure-on-contribute-site.

Problem is that storing metadata happens in a separate thread, but we don't wait at all before checking its value. This can introduce a race-condition if the storing metadata process has not completed before we check it, giving us random test failures.

This change modifies the acceptance tests - wrapping the method that checks the storeMetadata process in an Eventually with a timeout of 60 seconds and an interval of 1 second.
This will check every second for the storeMetadata value to return for 60 seconds.

Acceptance tests checked and ran successfully.

### DIFF
--- a/test/controllers/PaypalControllerSpec.scala
+++ b/test/controllers/PaypalControllerSpec.scala
@@ -22,14 +22,15 @@ import services.{ContributionOphanService, PaymentServices, PaypalService}
 import cats.instances.future._
 import com.gu.acquisition.typeclasses.AcquisitionSubmissionBuilder
 import org.mockito.internal.verification.VerificationModeFactory
-import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.concurrent.{Eventually, ScalaFutures}
+import org.scalatest.time.{Millis, Seconds, Span}
 import play.api.libs.json.Json
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.{ExecutionContext, Future}
 import scala.reflect.ClassTag
 
-class PaypalControllerFixture(implicit ec: ExecutionContext) extends MockitoSugar {
+class PaypalControllerFixture(implicit ec: ExecutionContext) extends MockitoSugar with Eventually {
 
   val mockPaymentServices: PaymentServices = mock[PaymentServices]
 
@@ -46,6 +47,8 @@ class PaypalControllerFixture(implicit ec: ExecutionContext) extends MockitoSuga
   val mockOphanService = mock[ContributionOphanService]
 
   val supportConfig = SupportConfig("https://support.thegulocal.com/thankyou")
+
+
 
   Mockito.when(mockPaymentServices.paypalServiceFor(Matchers.any[Request[_]]))
     .thenReturn(mockPaypalService)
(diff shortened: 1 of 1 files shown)


michaelwmcnamara:
So currently this will check for 60 seconds or until the expected value is returned.
If the expected value is returned, it will stop checking and return a result.
 If the value returned is not the expected one then this will keep checking until the timeout.
This means that a failing check will take 60 seconds to complete. 
On the other hand we have a full minute for the storeMetadata method to return a value, so the test should be much more resilient. Interested to hear folks opinion on this. Is a 60 second timeout too long?

test/controllers/PaypalControllerSpec.scala
```diff
@@ -47,6 +48,8 @@ class PaypalControllerFixture(implicit ec: ExecutionContext) extends MockitoSuga
 
   val supportConfig = SupportConfig("https://support.thegulocal.com/thankyou")
 
+
```
jacobwinch: Looks like we're accidentally adding whitespace here

test/controllers/PaypalControllerSpec.scala
```diff
@@ -223,7 +228,7 @@ class PaypalControllerSpec extends PlaySpec
       val result: Future[Result] = Helpers.call(fixture.controller.capturePayment, captureRequest)
       status(result).mustBe(200)
 
-      fixture.numberOfCallsToStoreMetaDataMustBe(1)
+      Eventually.eventually(timeout(Span(5, Seconds)), interval(Span(50, Millis))) {fixture.numberOfCallsToStoreMetaDataMustBe(1)}
```
jacobwinch: Do we need to wrap this in an eventually too? I'd have thought that using the eventually in numberOfCallsToStoreMetaDataMustBe would be sufficient.

test/controllers/PaypalControllerSpec.scala
```diff
@@ -80,22 +83,24 @@ class PaypalControllerFixture(implicit ec: ExecutionContext) extends MockitoSuga
...
-      captor[Option[String]],
-      captor[Option[String]],
-      captor[Option[String]],
-      captor[Option[String]],
-      captor[Option[IdentityId]],
-      captor[Option[String]],
-      captor[Option[String]]
-    )(captor[LoggingTags])
+    eventually(timeout(Span(60, Seconds)), interval(Span(1, Seconds))) {
+      def captor[A <: AnyRef](implicit classTag: ClassTag[A]): A =
```
jacobwinch: Personally I'd find it slightly easier to follow this if the captor function was declared outside of the eventually

test/controllers/PaypalControllerSpec.scala
michaelwmcnamara: good catch!

test/controllers/PaypalControllerSpec.scala
michaelwmcnamara: Hi Jacob, thats a good idea! 
Implementing now.

test/controllers/PaypalControllerSpec.scala
```diff
@@ -79,23 +80,25 @@ class PaypalControllerFixture(implicit ec: ExecutionContext) extends MockitoSuga
 
   val controller: PaypalController = new PaypalController(mockPaymentServices, mockCorsConfig, supportConfig, mockCsrfCheck, mockCloudwatchMetrics, mockOphanService)
 
+  def captor[A <: AnyRef](implicit classTag: ClassTag[A]): A =
+    ArgumentCaptor.forClass[A](classTag.runtimeClass.asInstanceOf[Class[A]]).capture()
+
   def numberOfCallsToStoreMetaDataMustBe(times: Int): Unit = {
```
jacobwinch, line 86: Minor style point, but I think that the outer braces can be removed here

jacobwinch:
Looks good to me 👍

jacobwinch (approved):

PR #406, opened 2018-01-25
Change the return url for PayPal payments from support.theguardian.com

jranks123:
In https://github.com/guardian/support-frontend/pull/466, we add a new endpoint on `support.theguardian.com` to handle Paypal one-off contributions. 

Previously, when someone who came from `support.theguardian.com` clicked on the confirm payment button in the PayPal UI, we redirected them to `/paypal/{countryGroup}/execute` on `contribute.theguardian.com`  which executed the payment, then redirected the user to the thank you page on support.theguardian.com

The problem here was that we couldn't send the email along with the redirect to the thank you page, as we aren't allowed to send it as a `.theguardian.com` session cookie or as a query string parameter. This meant that it was impossible to send the email address to the thank you page, meaning we couldn't ask these users if they would like to opt in to marketing emails. 

To resolve this, rather than redirect people from Paypal -> contribute, which executes the payment then redirects to support, we redirect people from Paypal -> support, where we make a GET request on the `/paypal/{countryGroup}/execute` endpoint which, after execution, returns the email in the body of the response. 

You can see a before and after of this process here:

Before:
old

After
emailroute


conf/PROD.public.conf
```diff
@@ -46,8 +46,8 @@ cors {
     ]
 }
 
-support-frontend {
-    thank-you-url = "https://support.theguardian.com/contribute/one-off/thankyou"
+support-backend {
+    paypal-execute = "https://support.code.dev-theguardian.com/paypal/execute"
```
joelochlann: This should be pointing to prod!

app/controllers/PaypalController.scala
```diff
@@ -154,18 +154,20 @@ class PaypalController(paymentServices: PaymentServices, corsConfig: CorsConfig,
 
     def okResult(payment: Payment): Result = {
       val response = render {
-        case Accepts.Json() => Ok(JsNull)
+        case Accepts.Json() => {
```
joelochlann: you could write this as

```scala
case Accepts.Json() if supportRedirect.contains(true) => Ok(Json.obj("email" -> payment.getPayer.getPayerInfo.getEmail))
case Accepts.Json() => Ok(JsNull)
```

Just a stylistic point though so I'll leave it up to you to decide which you think is nicer

test/controllers/PaypalControllerSpec.scala
```diff
@@ -168,25 +168,27 @@ class PaypalControllerSpec extends PlaySpec
 
   "Paypal Controller" should {
 
-    "generate correct redirect URL for support's successful PayPal payments" in {
+    "return the email address when called from support (and accept type is application/json)" in {
```
joelochlann, line 171: This test is nice - very clear. Maybe we could we add one to test when it's called from contribute?

app/services/PaypalService.scala
```diff
@@ -50,7 +49,8 @@ class PaypalService(
   config: PaypalApiConfig,
   contributionData: ContributionData,
   identityService: IdentityService,
-  emailService: EmailService
+  emailService: EmailService,
+  supportPaypalExecuteEndpoint: String
```
joelochlann, line 53: I guess it would be nice if you could just pass in the baseReturnUrl, rather than have the support endpoint passed in but the contribute endpoint fetched from config within the `PaypalService`. Though having looked at how baffling the config is ([this line](https://github.com/guardian/contributions-frontend/blob/34bef65b8825c2b413c18857d9001b9db2c4a8f8/test/resources/application.conf#L41-L42) in particular had me scratching my head... do we even use Jenkins!??) I can get why didn't want to change any of this. And this codebase is going away soon...

joelochlann (changes requested):
Great work with this and the other PR on support-frontend. Especially love the diagrams! 💎 

Some very minor comments which I'll leave up to you to decide about. Just one blocker though, `PROD.public.conf`

conf/PROD.public.conf
jranks123: Thanks for spotting this!

test/controllers/PaypalControllerSpec.scala
jranks123: This test already exists below, in the `"generate correct redirect URL for successful PayPal payments"` I'll make the name clearer

app/services/PaypalService.scala
jranks123: Yep I'm gonna leave this one for now.

app/controllers/PaypalController.scala
jranks123: I prefer your style, I'll go for it

joelochlann (approved):
🚢  it!

Mullefa:
BEAUTIFUL diagrams 😍😍

PR #407, opened 2018-01-26
Update to terms and conditions max width, update yellow on pledged so…

jerseywonder:
… far bar, update links colours


joelochlann (approved):
Looks good!

PR #410, opened 2018-02-02
Update current and new contributions databases at once

jranks123:
We are migrating our contributions store from PostGres to Aurora. Step one of this migration process is to update both the current DB and the new DB. This PR implements this.

The only difference in inserting into the new Aurora database is that custom types have to be referenced by `public.<custom type>`, rather than just `<custom type>`. This is why we have to instantiate the ContributionDataInstance with `public.paymentprovider` and `public.paymentstatus` for the Aurora instance, and  just `paymentprovider` and `paymentstatus` for the current postgres database

Once we have estabilshed that the new instance is working correctly, we can revert to just updating one database.

This change involves a change to the config. The `db` config will now have an extra config called [TEST/LIVE]_aurora, which contains the config of the new database.

Therefore, before merging, the config files for CODE, PROD and DEV need updating. As this config change is additive, it is a non-breaking change (i.e the current build of master works with the new config) 

Big shout out to @Mullefa who suggested I implement this with traits, which made the solution far more elegant

### DIFF
--- a/app/data/ContributionData.scala
+++ b/app/data/ContributionData.scala
@@ -14,18 +14,44 @@ import play.api.db.Database
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
 
-class ContributionData(db: Database)(implicit ec: ExecutionContext) extends TagAwareLogger {
+trait ContributionData {
+  def insertPaymentHook(paymentHook: PaymentHook)(implicit tags: LoggingTags): EitherT[Future, String, PaymentHook]
+  def insertPaymentMetaData(pmd: ContributionMetaData)(implicit tags: LoggingTags): EitherT[Future, String, ContributionMetaData]
+  def saveContributor(contributor: Contributor)(implicit tags: LoggingTags): EitherT[Future, String, Contributor]
+}
+
+class CurrentAndNewContributionData(currentDB: ContributionDataInstance, newDB: ContributionDataInstance) extends ContributionData {
+
+  override def insertPaymentHook(paymentHook: PaymentHook)(implicit tags: LoggingTags): EitherT[Future, String, PaymentHook] = {
+    newDB.insertPaymentHook(paymentHook)
+    currentDB.insertPaymentHook(paymentHook)
+  }
+
+  override def insertPaymentMetaData(pmd: ContributionMetaData)(implicit tags: LoggingTags): EitherT[Future, String, ContributionMetaData] = {
+    newDB.insertPaymentMetaData(pmd)
+    currentDB.insertPaymentMetaData(pmd)
+  }
+
+  override def saveContributor(contributor: Contributor)(implicit tags: LoggingTags): EitherT[Future, String, Contributor] = {
+    newDB.saveContributor(contributor)
+    currentDB.saveContributor(contributor)
+  }
+}
+
+class ContributionDataInstance(db: Database)(implicit ec: ExecutionContext) extends ContributionData with TagAwareLogger {
 
   def withAsyncConnection[A](autocommit: Boolean = false)(block: Connection => A)(implicit tags: LoggingTags): EitherT[Future, String, A] = EitherT(Future {
     val result = Try(db.withConnection(autocommit)(block))
     Either.fromTry(result).leftMap { exception =>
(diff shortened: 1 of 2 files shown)


app/wiring/AppComponents.scala
```diff
@@ -43,10 +43,15 @@ trait AppComponents extends PlayComponents with GzipFilterComponents {
     Cookies.authProvider(identityKeys).withDisplayNameProvider(Token.authProvider(identityKeys, "membership"))
 
   val contributionDataPerMode: Map[PaymentMode, ContributionData] = {
+    val postgresConfigId = "postgres"
```
Mullefa: don't think this is being used

app/wiring/AppComponents.scala
```diff
@@ -43,10 +43,15 @@ trait AppComponents extends PlayComponents with GzipFilterComponents {
...
+    val auroraConfigId = "aurora"
```
Mullefa: this is only be used in one place - perhaps it can just be included in the string that is being interpolated using it?

app/wiring/AppComponents.scala
```diff
@@ -43,10 +43,15 @@ trait AppComponents extends PlayComponents with GzipFilterComponents {
...
     val dbConfig = config.getConfig("dbConf")
     def contributionDataFor(mode: PaymentMode) = {
-      val modeKey = dbConfig.getString(mode.entryName.toLowerCase)
-      new ContributionData(dbApi.database(modeKey))(jdbcExecutionContext) // explicit execution context to avoid blocking the app
+      val postgresModeKey = dbConfig.getString(s"${mode.entryName.toLowerCase}")
```
Mullefa: don't need to interpolate the string here

app/data/ContributionData.scala
```diff
@@ -14,7 +14,31 @@ import play.api.db.Database
...
+  def insertPaymentHook(paymentHook: PaymentHook)(implicit tags: LoggingTags): EitherT[Future, String, PaymentHook]
+  def insertPaymentMetaData(pmd: ContributionMetaData)(implicit tags: LoggingTags): EitherT[Future, String, ContributionMetaData]
+  def saveContributor(contributor: Contributor)(implicit tags: LoggingTags): EitherT[Future, String, Contributor]
+}
+
+class CurrentAndNewContributionData(currentDB: ContributionDataInstance, newDB: ContributionDataInstance) extends ContributionData {
+
+  override def insertPaymentHook(paymentHook: PaymentHook)(implicit tags: LoggingTags): EitherT[Future, String, PaymentHook] = {
+    currentDB.insertPaymentHook(paymentHook)
+    newDB.insertPaymentHook(paymentHook)
```
Mullefa: the `currentDB` result should probably be returned - just in case the result is being used anywhere - likewise for the other overridden methods in this class

app/data/ContributionData.scala
```diff
@@ -14,7 +14,31 @@ import play.api.db.Database
...
+    newDB.insertPaymentMetaData(pmd)
+  }
+
+  override def saveContributor(contributor: Contributor)(implicit tags: LoggingTags): EitherT[Future, String, Contributor] = {
+    currentDB.saveContributor(contributor)
+    newDB.saveContributor(contributor)
+  }
+}
+
+class ContributionDataInstance(db: Database)(implicit ec: ExecutionContext) extends ContributionData with TagAwareLogger {
```
Mullefa: if the database name could be accessed (e.g. by an additional parameter in the class constructor, or a method on the `Database` class (?)), then we could reference this in the log messages. This may prove useful for debugging purposes - especially as the calls are asynchronous which means the data base can not be inferred from the order of the log messages.

Mullefa (commented):
couple of points - looks good overall - props for working through the ambiguous config setup! 😅

app/data/ContributionData.scala
jranks123: database has a `name` property which, in this case, will be [TEST/LIVE] or [TEST/LIVE]_auroroa. Good shout on the logging

Mullefa: 👌

PR #415, opened 2018-05-03
Redirect to point Support platform 

svillafe:
To point the contribute pages to support.


jranks123 (approved):

svillafe:
Since / has a redirect now, Prout can't pick up that is deployed.

